PR #5, opened 2017-12-04
V1: use pure ts style code

fradinni:
- remove functions in callbacks
- add "prefer-const" in tslint
- update README
- Update NodeJS version in package.json from ^6.2.2 to >=6.2.2 <9.0.0


cfsnyder:
thanks @fradinni !

PR #14, opened 2018-01-29
Add catalog/vappTemplate/Media support and add option to define the a…

JoMas971:
Add catalog/vappTemplate/Media support and add option to define the Api/KC urls in env.

### DIFF
--- a/src/sdk/basic-configuration.ts
+++ b/src/sdk/basic-configuration.ts
@@ -1,0 +1,28 @@
+/**
+ * Basic configurations for SDK using environment variables.
+ */
+export class BasicConfiguration {
+  /**
+   * Get the default API url.
+   * @returns {string}
+   */
+  static getApiUrl(): string {
+    if (process.env.ILAND_API_URL && process.env.ILAND_API_URL !== 'undefined') {
+      return process.env.ILAND_API_URL;
+    } else {
+      return 'https://api.ilandcloud.com';
+    }
+  }
+
+  /**
+   * Get the default authorisation url.
+   * @returns {string}
+   */
+  static getAuthorizationUrl(): string {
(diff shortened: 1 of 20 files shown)


src/sdk/basic-configuration.ts
```diff
@@ -0,0 +1,28 @@
...
+    } else {
+      return 'https://api.ilandcloud.com';
+    }
+  }
+
+  /**
+   * Get the default keycloak authorisation url.
+   * @returns {string}
+   */
+  static getKeycloakUrl(): string {
```
cfsnyder: Lets call it `getAuthorizationUrl()` and lets update the environment variable, etc to be generic as well

src/sdk/model/__tests__/vapp-template.test.unit.ts
```diff
@@ -0,0 +1,61 @@
...
+beforeAll(() => {
+  Iland.init(new IlandDirectGrantAuthProvider({
+    username: '',
+    password: '',
+    clientSecret: '',
+    clientId: ''
+  }));
+});
+
+function runMediaAssertionsAgainsMock(vappTemplate: VappTemplate) {
```
cfsnyder: misnamed

src/sdk/basic-configuration.ts
JoMas971: Ok let me change the name though.

src/sdk/model/__tests__/vapp-template.test.unit.ts
JoMas971: Nice catch thanks ! Will update this.

JoMas971:
Should be good now. I've fixed your comments.

PR #15, opened 2018-02-13
Add IAM service to control user permissions and cleanup.

JoMas971:


src/sdk/model/user.ts
```diff
@@ -41,6 +44,33 @@ export class User {
     });
   }
 
+  /**
+   * Get the security for the specified user and store them into the user properties.
+   * It would actually get the list of companyInventory and roles for the specified user.
+   * Be careful when using this method, it should be called only once since it can be slow.
+   * @returns {Promise<User>}
+   */
+  async getUserWithSecurity(): Promise<User> {
```
cfsnyder: I thought we agreed to create a `UserWithSecurity` type to extend the `User` type with the `isUserPermitted`

src/sdk/service/iam-service.ts
```diff
@@ -0,0 +1,178 @@
...
+    if (permissionObject) {
+      // handle public catalogs, media, and templates
+      switch (permissionObject.domain) {
+        case 'ILAND_CLOUD_CATALOG':
+          return (entityUuid.includes('public:') && permission === 'VIEW_ILAND_CLOUD_CATALOG');
+        case 'ILAND_CLOUD_MEDIA':
+          return (entityUuid.includes('public:') && (permission === 'VIEW_ILAND_CLOUD_MEDIA' ||
+            permission === 'CLONE_DOWNLOAD_ILAND_CLOUD_MEDIA'));
+        case 'ILAND_CLOUD_VAPP_TEMPLATE':
+          return (entityUuid.includes('public:') && (permission === 'VIEW_ILAND_CLOUD_VAPP_TEMPLATE' ||
```
cfsnyder: You shouldn't be checking individual permissions on these. Instead you should be checking whether the specified permission's AccessType is READ

src/sdk/service/iam-service.ts
```diff
@@ -0,0 +1,178 @@
...
+      IamService.instance = new IamService();
+    }
+    return IamService.instance;
+  }
+
+  /**
+   * Get the permission service.
+   * @returns {PermissionService}
+   */
+  get permissionService(): PermissionService {
```
cfsnyder: is there a good reason to have this here? Seems more natural and less confusing to just use PermissionService.getInstance() wherever we need it

src/sdk/model/user.ts
JoMas971: Yeah that was the first thing we said, but finally, you told me that maybe it would be ok to add this to the actual user class.
"after the user logs in, in the console we do:
`User.getCurrentUser().then(function(user) { user.getUserWithSecurity().then(function(userWithSecurity) { $rootScope.user = userWithSecurity; })})` (edited)
and we only load all of the roles/inventory when `getUserWithSecurity` is called"

src/sdk/service/iam-service.ts
JoMas971: Yeah ok. I can do this. np

src/sdk/service/iam-service.ts
JoMas971: Ok. I'll update this.

src/sdk/model/permission-map.ts
```diff
@@ -0,0 +1,288 @@
...
+        false, ['MANAGE_ILAND_CLOUD_VM_POWER_STATE'])));
+    this._permissions.set('MANAGE_ILAND_CLOUD_VAPP_CONFIGURATION',
+      (new Permission('MANAGE_ILAND_CLOUD_VAPP_CONFIGURATION', 'ILAND_CLOUD_VAPP', 'WRITE', true,
+        false, null)));
+    this._permissions.set('MANAGE_ILAND_CLOUD_VAPP_SNAPSHOTS',
+      (new Permission('MANAGE_ILAND_CLOUD_VAPP_SNAPSHOTS', 'ILAND_CLOUD_VAPP', 'WRITE', true,
+        false, null)));
+    this._permissions.set('MANAGE_ILAND_CLOUD_VAPP_LEASES',
+      (new Permission('MANAGE_ILAND_CLOUD_VAPP_LEASES', 'ILAND_CLOUD_VAPP', 'WRITE', true,
+        false, ['MANAGE_ILAND_CLOUD_VM_SNAPSHOTS'])));
```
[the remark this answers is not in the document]
JoMas971: Ok i'll do. Have you updated this core side already ? Because this is what i've got from the permissions java file you give me.

src/sdk/model/company-inventory.ts
```diff
@@ -79,31 +80,53 @@ export class CompanyInventory {
...
+      return undefined;
+    }
     return this._uuidMap[uuid];
   }
 
+  /**
+   * Get all entities mapped by their types.
+   * @returns {{[p: string]: Array<InventoryEntity>}}
+   */
+  getAllEntities(): { [type: string]: Array<InventoryEntity> } {
```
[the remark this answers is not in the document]
JoMas971: ok, i'll do.

src/sdk/model/user.ts
JoMas971: ok i see your point. I'll do it.

PR #16, opened 2018-02-15
Add IAM service to control user permissions and cleanup.

JoMas971:


src/sdk/__mocks__/http.ts
```diff
@@ -71,6 +123,14 @@ export class Http {
 
   async get(url: string, config?: AxiosRequestConfig): Promise<AxiosResponse> {
     switch (true) {
+      case /always\/unauthorized/.test(url):
```
cfsnyder: I thought we agreed to remove these because they doesn't actually test any of our code, it just exercises mocks

PR #18, opened 2018-02-27
Add getRole function to User class and fix the node engine

JoMas971:


JoMas971:
@cfsnyder i've moved the `getRole()` function to the user class since we need it to get the user role for a specific company. But i've kept the rest separate. I'll let you merge this if it's ok for you ;-).

src/sdk/model/user.ts
```diff
@@ -224,6 +225,18 @@ export class User {
     });
   }
 
+  /**
+   * Gets the user's role for a company
+   * @param {string} companyUuid
+   * @returns {Promise<Role>}
+   */
+  async getRoleFor(companyUuid: string): Promise<Role> {
```
cfsnyder: can we change the name to `getRole`

JoMas971: sure. let me do that

JoMas971:
@cfsnyder it should be good now ;-).

cfsnyder:
It has conflicts @JoMas971

JoMas971:
Should be good now @cfsnyder

PR #24, opened 2018-03-08
Update role, policy and IAM Service.

JoMas971:


src/sdk/service/iam-service.ts
```diff
@@ -151,4 +157,97 @@ export class IamService {
...
+        errors.push((new Error('Entity not found in this company.')));
+      } else {
+        if (policy.entityDomain !== entityOptional.type) {
+          errors.push((new Error('The entity is assigned in policy but not for the good domain.')));
+        }
+        if (policy.type === 'CUSTOM') {
+          if (policy.permissions.length === 0) {
+            errors.push((new Error('Custom policies must contain at least one permission.')));
+          } else {
+            // for (const p of policy.permissions) {
```
JoMas971: Oops! I'm updating this. It shouldn't be there...

src/sdk/model/policy.ts
```diff
@@ -74,14 +85,77 @@ export class PolicyBuilder {
...
+    this._entityDomain = entityDomainType;
+    return this;
+  }
+
+  /**
+   * Set policy type.
+   * @param {PolicyType} type
+   * @returns {PolicyBuilder}
+   */
+  setPolicyType(type: PolicyType): PolicyBuilder {
```
cfsnyder: it would be better to not have these new setters because you're opening the door to missing validation issues. For example, imagine that a bunch of permissions are added for the VM domain and then the entity domain type is later changed to VAPP before being submitted.

src/sdk/service/iam-service.ts
```diff
@@ -13,7 +18,8 @@ export class IamService {
    * @param {Role} role
    * @returns {Policy | null}
    */
-  static getEffectivePolicy(companyInventory: CompanyInventory, entity: InventoryEntity, role: Role): Policy | null {
+  static getEffectivePolicy(companyInventory: CompanyInventory, entity: InventoryEntity,
```
cfsnyder: should be private

src/sdk/service/iam-service.ts
```diff
@@ -69,8 +75,8 @@ export class IamService {
    * @param {Role} role
    * @returns {Policy | null}
    */
-  static findFirstRelevantPolicy(companyInventory: CompanyInventory,
-                                 entity: InventoryEntity | undefined, role: Role): Policy | null {
+  static findFirstRelevantPolicy(companyInventory: CompanyInventory, entity: InventoryEntity | undefined,
```
cfsnyder: should be private

src/sdk/service/iam-service.ts
cfsnyder: why is this comment still in here?

src/sdk/service/iam-service.ts
```diff
@@ -151,4 +157,97 @@ export class IamService {
...
+   */
+  static validateRole(roleCreationRequest: RoleCreationRequest, companyInventory: CompanyInventory) {
+    const errors: Array<Error> = [];
+    if (roleCreationRequest.policies.length === 0) {
+      errors.push((new Error('A role must have at least one policy.')));
+    }
+    for (const policy of roleCreationRequest.policies) {
+      const entityOptional: InventoryEntity | undefined = companyInventory.getEntityByUuid(policy.entityUuid);
+      if (entityOptional === undefined) {
+        errors.push((new Error('Entity not found in this company.')));
```
cfsnyder: need to include the name of the entity in these so they're useful to the user

src/sdk/service/iam-service.ts
```diff
@@ -151,4 +157,97 @@ export class IamService {
...
+      } else {
+        if (policy.entityDomain !== entityOptional.type) {
+          errors.push((new Error('The entity is assigned in policy but not for the good domain.')));
```
cfsnyder: change to "Policy for entity '${?}' has domain type '${?}' but entity is actually of type '${?}'".

src/sdk/service/iam-service.ts
```diff
@@ -151,4 +157,97 @@ export class IamService {
...
+        // own permissions on the same domain
+        for (const permission of policy.permissions) {
+          const perm = PermissionService.getPermission(permission);
+          if (perm && perm.impliedPermissions) {
+            for (const impliedPermission of perm.impliedPermissions) {
+              const implPerm = PermissionService.getPermission(impliedPermission);
+              if (implPerm && implPerm.domain === policy.entityDomain && !policy.hasPermission(impliedPermission)) {
+                // the policy must have this permission, otherwise it violates an
+                // implication of higher-level policies
+                errors.push((new Error('Policy must contain permission=' + impliedPermission +
```
cfsnyder: again, need to include name of entity for the policy so that the message is useful to a user

src/sdk/model/policy.ts
JoMas971: Humm ok. I'll remove them. It was just easier to use (specially for my tests) but you're right, it's not good.

src/sdk/service/iam-service.ts
JoMas971: We can't get the entity name but only the Entity UUID there. Because the entity doesn't exist so we can't get the name ;-). But i'll put the UUID then.

PR #27, opened 2018-03-26
Add Company Logo methods and tests

fradinni:


src/sdk/model/__tests__/company.test.unit.ts
```diff
@@ -175,3 +175,50 @@ test('Properly submits request to get users with a specific role', async() => {
...
+test('Properly submits request to get company logo', async() => {
+  const id = MockCompanyJson.uuid;
+  return Company.getCompany(id).then(async(company) => {
+    expect(Iland.getHttp().get).lastCalledWith(`/companies/${id}`);
+    expect(company.entityType).toBe('COMPANY');
+    return company.getLogo().then(async (logo) => {
+      expect(Iland.getHttp().get).lastCalledWith(
+        `/companies/${id}/logo`,
+        {
+          'headers': {'Accept': 'image/vnd.ilandcloud.api.v1.0+jpeg'},
```
cfsnyder: lets not hard-code the version in here because it will need to be updated every time we do a release.

src/sdk/model/company.ts
```diff
@@ -177,4 +177,32 @@ export class Company extends Entity {
...
+      },
+      responseType: 'arraybuffer'
+    }).then((response) => {
+      return new Uint8Array(response.data);
+    }, () => {
+      return undefined;
+    });
+  }
+
+  async deleteLogo(): Promise<boolean> {
```
cfsnyder: lets do Promise<any> and not map the result since the boolean value is meaningless

src/sdk/model/company.ts
```diff
@@ -177,4 +177,32 @@ export class Company extends Entity {
...
+        }
+    }).then(() => {
+        return true;
+    });
+  }
+
+  async getLogo(): Promise<Uint8Array|undefined> {
+    return Iland.getHttp().get(`/companies/${this.uuid}/logo`, {
+      headers: {
+        'Accept': 'image/vnd.ilandcloud.api.v1.0+jpeg'
```
cfsnyder: lets not hard-code the version in here. Versioning should be encapsulated in the http class

src/sdk/model/company.ts
```diff
@@ -177,4 +177,32 @@ export class Company extends Entity {
     });
   }
 
+  async setLogo(logo: Uint8Array): Promise<boolean> {
+    return Iland.getHttp().post(`/companies/${this.uuid}/logo`, logo,{
+        headers: {
+            'Content-Type': 'image/jpeg'
+        }
+    }).then(() => {
+        return true;
```
cfsnyder: lets use return type of Promise<any> and not do this since this boolean value is meaningless

src/sdk/model/__tests__/company.test.int.ts
```diff
@@ -113,3 +113,32 @@ test('Can refresh company', async() => {
     });
   });
 });
+
+test('Can set logo', async() => {
```
cfsnyder: if we're going to have integration tests for this, I'd prefer that we put back the company's original logo in the end. We have to remember that customers will be using this in the future and it would be annoying if it removes/updates their logo if they're running these tests.

src/sdk/http.ts
```diff
@@ -58,7 +58,14 @@ export class Http {
         if (str.indexOf(')]}\'\n') === 0) {
           str = str.substring(5);
         }
-        error = JSON.parse(str) as ApiErrorJson;
+        try {
+          error = JSON.parse(str) as ApiErrorJson;
+        } catch(e) {
+          error = {
+            message: reason.response!!.status + " " + reason.response!!.statusText
```
cfsnyder: is this for a 404 case with your new endpoint?

fradinni: Yeah when company logo is not set it returns an error ans it was failing at this point.

fradinni:
PR updated

Issue #31, opened 2018-04-05
Add performance sample endpoints to all entities


cfsnyder:
We need to add our performance sample endpoints at all levels.

PR #34, opened 2018-04-06
#31: Add perf sample endpoint for all entities

fradinni:


src/sdk/model/entity.ts
```diff
@@ -55,4 +58,40 @@ export abstract class Entity {
    */
   abstract get entityType(): EntityType;
 
+  /**
+   * Returns the api prefix for current entity or null if not defined
+   * @returns {string} the api prefix or null
+   */
+  get apiPrefix(): string | null {
```
cfsnyder: I really don't like this. If the interface method doesn't apply to every entity then it shouldn't be a method on the interface. It kinda defeats the purpose of having the interface in the first place. I also don't like having this apiPrefix method, its  detail that should be abstracted away by this SDK.

src/sdk/model/json/perf-samples.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * API PerfSampleSerie JSON Representation.
+ */
+export interface PerfSampleSerie {
```
cfsnyder: all of our current JSON interface representations have names that end in `Json`, lets not deviate from that naming precedent.

PR #40, opened 2018-04-06
31: Add perf samples endpoints

fradinni:
PR for issue #31

### DIFF
--- a/src/sdk/model/mixins/perf-samples/_json_/perf-samples-request.ts
+++ b/src/sdk/model/mixins/perf-samples/_json_/perf-samples-request.ts
@@ -1,0 +1,12 @@
+/**
+ * Role Creation Request JSON properties.
+ */
+import { PerfCounterJson } from './perf-samples';
+
+export interface PerfSamplesRequestJson {
+  counter: PerfCounterJson;
+  start: number;
+  end: number;
+  interval: number;
+  limit: number;
+}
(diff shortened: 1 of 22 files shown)


src/sdk/model/mixins/entity-with-perf-samples.ts
```diff
@@ -0,0 +1,31 @@
...
+  uuid: string;
+  apiPrefix: string;
+
+  async getPerfCounters(): Promise<Array<PerfCounterJson>> {
+    return Iland.getHttp().get(`${this.apiPrefix}/${this.uuid}/performance-counters`).then((response) => {
+      return response.data as Array<PerfCounterJson>;
+    });
+  }
+
+  async getPerfSamples(request: PerfSamplesRequestJson): Promise<PerfSampleSerieJson> {
```
cfsnyder: One last thing on these @fradinni . You shouldn't be returning the Json interface responses at this layer, they should be wrapped by corresponding classes. We're doing this so that we have a way to provide convenience methods associated with the data, to convert Json types to more useful javascript types, and to provide a buffer between server response format and the client integration code so that we can have better control over when we need to break compatibility with future changes.

cfsnyder:
@fradinni I didn't notice before that you hadn't actually mocked up your responses and tested each of the model methods. I'll need you to do that before I merge so that our code coverage doesn't dip down. See the codecov report to understand all of the lines that aren't yet tested.

cfsnyder:
Thanks @fradinni

PR #42, opened 2018-04-09
Add edges GET endpoints + tests

JoMas971:


src/sdk/auth/direct-grant-auth-provider.ts
```diff
@@ -1,14 +1,25 @@
...
 export class IlandDirectGrantAuthProvider implements AuthProvider {
 
   private _token: Token|undefined;
+  private _tokenObservable: Observable<string>;
 
   constructor(private _config: IlandDirectGrantAuthConfig) {
+    this._tokenObservable = Observable.create((observable: Subscriber<string>) => {
+      this.getToken().then(token => {
+        observable.next(token);
+        observable.complete();
```
cfsnyder: why are you completing it here? Shouldn't you be continually broadcasting tokens in the refresh method or am I missing something?

src/sdk/http.ts
```diff
@@ -106,7 +106,7 @@ export class Http {
    * @param {string} url the URL path
    * @param {AxiosRequestConfig} config request configuration
    * @returns {Promise<AxiosResponse>} promise that resolves with server response
-   * @throws {ApiError} if the server responds with an error
+   * @throws {ApiError} if the server responds with an erroro some maintenance on the core right now?
```
cfsnyder: lets revert that, doesn't really make sense

src/sdk/model/edge.ts
```diff
@@ -144,6 +158,217 @@ export class Edge extends Entity {
     });
   }
 
+  /**
+   * Get the Edge dhcp pool
+   * @returns {Promise<EdgeDhcpJson>}
+   */
+  async getDhcp(): Promise<EdgeDhcpJson> {
```
cfsnyder: Why are you not wrapping all of the *Json interfaces with corresponding classes like we've done with everything else in this SDK?

src/sdk/auth/direct-grant-auth-provider.ts
JoMas971: Well, the `direct-grant-auth-provider` is only used for tests. It will automatically return a connected user/token. There, it doesn't use the keycloak lib at all. It uses a regular Axios request. But because of this, i don't have the event listener to refresh the token.
But, if you think we need to have that to this class, i could create an interval that call the `getToken`method every x seconds. But i don't think it worth it in this particular case.

src/sdk/http.ts
JoMas971: Wow, good catch. Don't know why my IDE did it. Doesn't make sense indeed.

src/sdk/model/edge.ts
JoMas971: I don't understand what you're talking about :-/.

JoMas971: Ho you mean i should create a ts class instead of only getting the Json interface ? Because a class is not needed and not even usefull in our case. We don't need to instanciate a class. We only need to get the value from the API. If we needed to do object manipulation we would have create a class but there, it's not needed. When we would add the `POST`, `PUT` and `DELETE` parts, we would create a builder class.

src/sdk/auth/direct-grant-auth-provider.ts
cfsnyder: Its not only used for tests, its used for any nodejs client. So yes, we need it to work properly

src/sdk/model/edge.ts
cfsnyder: I wouldn't say that its not useful. Some good reasons to continue doing this are: it allows us to provide convenience methods associated with the data (e.g. we could add a method to get an array of all IPs on an edge inteface), to convert Json types to more useful javascript types (e.g. to convert a number representing epoch millis from the API to a javascript date), and to provide a buffer between server response format and the client integration code so that we can have better control over when we need to break compatibility with future changes. I would definitely like to continue with this pattern that we've been using thus far.

src/sdk/auth/direct-grant-auth-provider.ts
JoMas971: Ok. I'll do the interval solution though. I was thinking of doing one request every seconds. Do you think it makes sense ?

src/sdk/model/edge.ts
JoMas971: ok i see you point. Well. I'll create classes for all of this JSON object though.

src/sdk/auth/direct-grant-auth-provider.ts
cfsnyder: there is no need to user an interval.. you can emit from the observer from within the _refresh method since it controls the token refresh exclusively

JoMas971: You mean that i just need to emit an event everytime we call the refresh method ?! Yeah i can do it. Better than an interval...

JoMas971: Finally i've decided to do like KC (create an `onRefreshToken` method within the class, and assign this method tomy observer and call `onRefreshToken()`from the `_refreshToken()` method. :-)

PR #45, opened 2018-04-11
add get cloud-tenant endpoint

ariquinones:


src/sdk/model/json/cloud-tenant.ts
```diff
@@ -0,0 +1,61 @@
...
+  contract_uuid: string;
+  location_id: string;
+  endpoint: string;
+  backup_count: number;
+}
+
+/**
+ * Interface for VeeamResource JSON properties.
+ */
+export interface VeeamResource {
```
cfsnyder: all of these interfaces that describe JSON formats should have names that end in `Json`

src/sdk/model/cloud-tenant.ts
```diff
@@ -0,0 +1,166 @@
...
+   */
+  get enabled(): boolean {
+    return this._json.enabled;
+  }
+
+  /**
+   * Gets the VeeamResource of the Cloud Tenant.
+   * @returns {VeeamResource} value
+   */
+  get resources(): VeeamResource {
```
cfsnyder: There is no reason to make the javascript/typescript representation match the representation in the JSON. In fact, one of the main reasons for writing these wrapper classes is to make the Javascript/Typescript client interface better than the interface provided by the API. This is a perfect case actually.

ariquinones: @cfsnyder pr updated

cfsnyder:
Thanks @ariquinones !

PR #50, opened 2018-04-25
add vpg endpoints

ariquinones:
issue #50

### DIFF
--- a/src/sdk/model/vpg/json/vpg-failover-create-request.ts
+++ b/src/sdk/model/vpg/json/vpg-failover-create-request.ts
@@ -1,0 +1,20 @@
+/**
+ * Vpg Failover Create Request JSON properties.
+ */
+export interface VpgFailoverCreateRequestJson {
+  checkpoint_identifier: string;
+  commit_policy: CommitPolicy;
+  shutdown_policy: ShutdownPolicy;
+  time_to_wait_before_shutdown_in_sec: number;
+  commit_value: number;
+}
+
+/**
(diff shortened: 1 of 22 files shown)


JoMas971:
This PR looks fine to me. Good job man

src/sdk/model/vpg/json/vpg-failover-create-request.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Vpg Failover Create Request properties
+ */
+export interface VpgFailoverCreateRequest {
```
cfsnyder: Names of all interfaces in the `json` folder should end in 'Json', and all should be wrapped by a class at the SDK level. Use Nico's pattern for request classes that we talked about as a group.

src/sdk/model/vpg/json/vpg-failover-test-alert-request.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Vpg Failover Test Alert Request
+ */
+export interface VpgFailoverTestAlertRequest {
```
cfsnyder: another one that needs to be renamed and wrapped by a class.

src/sdk/model/vpg/json/vpg-failover-create-request.ts
JoMas971: Ho yeah. Good catch !

cfsnyder:
thanks @ariquinones !

PR #52, opened 2018-04-25
vdc metadata endpoints

ariquinones:
issue #52


src/sdk/model/vdc.ts
```diff
@@ -263,4 +266,50 @@ export class Vdc extends Entity {
...
+      });
+    });
+  }
+
+  /**
+   * Updates the vDC metadata.
+   * @param {Array<MetadataJson<MetadataType>>} metadata the new array of metadata
+   * @returns {Promise<Task>} task promise
+   */
+  async updateMetadata(metadata: Array<MetadataJson<MetadataType>>): Promise<Task> {
```
cfsnyder: shouldn't be using a Json interface as part of a method definition at this level

Issue #61, opened 2018-05-22
Refactor org endpoints for getting descendant entities in association with API change


cfsnyder:
Array responses have been replaced with wrapper objects responses.

PR #62, opened 2018-05-22
#61: refactor org endpoints in association with change of array response format

cfsnyder:


src/sdk/model/org/__tests__/org.test.int.ts
```diff
@@ -96,3 +96,51 @@ test('Can refresh Org', async() => {
     });
   });
 });
+
+test('Can get org edges', async() => {
+  const org = await Org.getOrg(inventoryOrg.uuid);
```
JoMas971: It can be a good idea to add a `expect.assertions(num)` when num is the actual number of test expected. Specially when we have promise inside promise.

JoMas971: Nevermind. This is not needed there.

PR #64, opened 2018-05-23
add org historical billing endpoint

cfsnyder:


src/sdk/model/org/__tests__/org.test.int.ts
```diff
@@ -96,3 +96,15 @@ test('Can refresh Org', async() => {
     });
   });
 });
+
+test('Can get org historical billing', async() => {
+  return Org.getOrg(inventoryOrg.uuid).then(async function(org) {
```
JoMas971: It can be a good idea to add a `expect.assertions(num)` when num is the actual number of test expected. Specially when we have promise inside promise.

JoMas971: Nevermind. This is not needed there.

PR #67, opened 2018-05-30
#66 Add full role management to user class

JoMas971:
#66

### DIFF
--- a/src/sdk/model/user/user.ts
+++ b/src/sdk/model/user/user.ts
@@ -230,6 +230,17 @@ export class User {
   }
 
   /**
+   * Gets the user's list of companies.
+   * @returns {Promise<Array<Company>>} user's list of companies
+   */
+  async getCompanies(): Promise<Array<Company>> {
+    return Iland.getHttp().get(`/users/${this.username}/companies`).then((response) => {
+      const companiesList = response.data.data as Array<CompanyJson>;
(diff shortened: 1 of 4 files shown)


src/sdk/model/user/user.ts
```diff
@@ -242,14 +253,24 @@ export class User {
...
+      role_uuid: roleUuid
     });
   }
 
+  /**
+   * Un-assign a role from a user.
+   * @param {string} companyUuid
+   * @returns {Promise<any>}
+   */
+  async unAssignRole(companyUuid: string): Promise<any> {
```
cfsnyder: unassignRole

src/sdk/model/user/user.ts
```diff
@@ -229,6 +229,17 @@ export class User {
     });
   }
 
+  /**
+   * Gets the user's list of companies.
+   * @returns {Promise<Array<Company>>} user's list of companies
+   */
+  async getCompanies(): Promise<Array<Company>> {
+    return Iland.getHttp().get(`/users/${this.username}/companies`).then((response) => {
+      const companiesList = response.data as Array<CompanyJson>;
```
cfsnyder: is this one not refactored so that it should be `data.data` yet?

cfsnyder:
thanks @JoMas971

Issue #73, opened 2018-05-30
Add org compliance reports endpoint methods


cfsnyder:
We need all methods for the org compliance report API.

PR #76, opened 2018-05-31
Task #73: Add org compliance reports endpoint methods

fradinni:


src/sdk/model/org/reports/anti-malware-over-time.ts
```diff
@@ -0,0 +1,49 @@
+/* istanbul ignore next: autogenerated */
```
cfsnyder: i think this has to be sitting on top of the class declaration (not before the imports). you can verify but i believe i already tried that. comment applies to all classes

fradinni: ah yeah you right, let me update that

PR #88, opened 2018-06-06
Refactor http mock files

fradinni:


cfsnyder:
Nice @fradinni ! much cleaner like that

PR #91, opened 2018-06-06
Resolves #89: Add move endpoint for VM

### LINKED ISSUE #89: Add move endpoint for VM

fradinni:


src/sdk/model/vm/vm-copy-move-request.ts
```diff
@@ -0,0 +1,78 @@
+import { VmCopyMoveRequestJson, VmCopyMoveVmVnicRequestJson } from './__json__';
+import { VmCopyMoveVmVnicRequest } from './vm-copy-move-vm-vnic-request';
+
+/* istanbul ignore next: autogenerated */
+export class VmCopyMoveRequest {
+
+  private readonly _json: VmCopyMoveRequestJson;
```
JoMas971: Why read only for _json ? I guess it could be possible to only refresh the object so we would be able to update this prop.

src/sdk/model/vm/vm-copy-move-vm-vnic-request.ts
```diff
@@ -0,0 +1,87 @@
+import { IpAddressingMode, VmCopyMoveVmVnicRequestJson } from './__json__';
+
+/* istanbul ignore next: autogenerated */
+export class VmCopyMoveVmVnicRequest {
+
+  private readonly _json: VmCopyMoveVmVnicRequestJson;
```
JoMas971: Same here. Idon't think it should be read-only.

src/sdk/model/vm/vm-copy-move-request.ts
cfsnyder: it is never updated so it is correct to be marked read only

PR #119, opened 2018-06-08
Resolves #98: add batch failover endpoint methods

### LINKED ISSUE #98: Add batch failover endpoint methods

fradinni:


cfsnyder:
You only added one of the two endpoints @fradinni ?

PR #133, opened 2018-06-08
Update user class to add update and delete.

JoMas971:
#107

### DIFF
--- a/src/sdk/model/user/user.ts
+++ b/src/sdk/model/user/user.ts
@@ -7,6 +7,8 @@ import { UserType } from './__json__/user-type';
 import { UserInventoryJson } from './inventory-entity/__json__/user-inventory-json';
 import { RoleJson } from '../iam/role/__json__/role-json';
 import { CompanyJson } from '../company/__json__/company-json';
+import { UserDomainJson } from './__json__/user-domain-json';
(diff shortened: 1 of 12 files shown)


src/sdk/model/user/user.ts
```diff
@@ -199,6 +201,31 @@ export class User {
...
+  /**
+   * Delete the current user.
+   * @returns {Promise<void>}
+   * @throws Error
+   */
+  async delete(): Promise<void> {
+    return Iland.getHttp().delete(`/users/${this.username}`).then(() => {
+      return;
+    }).catch(() => {
+      throw new Error(`Something went wrong with user deletion.`);
```
cfsnyder: what is this?

JoMas971: Handling the case something went wrong with user delete API side. If it return a 4xx or 5xx request. The error message can be changed if you want.

cfsnyder: theres no reason to catch the error, we should let the api bubble up like we do everywhere else. It could give the user some interesting details about the reason. No reason for the success promise handler that just returns either.

cfsnyder:
thanks @JoMas971

PR #158, opened 2018-06-11
Resolves #132: Add VappNetworks get vm interfaces endpoint

### LINKED ISSUE #132: Add VappNetworks get vm interfaces endpoint

fradinni:


src/sdk/model/vapp-network/vapp-network-interface-list.ts
```diff
@@ -0,0 +1,35 @@
+import { VappNetworkInterfaceListJson } from './__json__';
+import { VappNetworkInterface } from './vapp-network-interface';
+
+/* istanbul ignore next: autogenerated */
+export class VappNetworkInterfaceList {
```
cfsnyder: I don't see you ever using this class (or the corresponding JSON interface)?

fradinni: Let me remove it, it's because We use an Array as return type in VappNetworks method so this class it not used in SDK.

fradinni:
@cfsnyder updated

PR #177, opened 2018-06-13
Resolves #150: Add vApp billing endpoint methods

### LINKED ISSUE #150: Add vApp billing endpoint methods

fradinni:
Also fixing other classes and test that were using old BillResponse format


cfsnyder:
thanks @fradinni !

PR #218, opened 2018-06-14
Resolves #193: Add vApp template endpoint methods for enabling downlo…

### LINKED ISSUE #193: Add vApp template endpoint methods for enabling download and getting download details

fradinni:
…ad and getting download details


fradinni:
@cfsnyder just rebased on master and fixed conflicts

src/sdk/model/vapp-template/vapp-template-download-details.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * vApp template Download Details.
+ */
+/* istanbul ignore next: autogenerated */
+import { VappTemplateDownloadDetailsJson } from './__json__/vapp-template-download-details-json';
```
cfsnyder: move this to the top and move tsdoc class comment down

src/sdk/model/vapp-template/vapp-template.ts
```diff
@@ -1,15 +1,17 @@
+import { Task } from '..';
```
cfsnyder: bad import

fradinni: I'll fix it. But... it's generated like this by IDE because Task is exported by index.ts file in the parent directory... So we'll have to check import when we are in this case.

fradinni:
Fixed.

PR #234, opened 2018-06-15
Resolves #208: Add edge endpoint for updating static routing service

### LINKED ISSUE #208: Add edge endpoint method for updating static routing service

nksprzak:


src/sdk/model/edge/edge.ts
```diff
@@ -30,6 +30,8 @@ import { EdgeFirewallLogType } from './firewall/__json__/edge-firewall-log-type'
 import { EdgeBackingConfigurationType } from './__json__/edge-backing-configuration-type';
 import { EdgeInterfaceJson } from './edge-interface/__json__/edge-interface-json';
 import { EntityType } from '../common/__json__/entity-type';
+import { Task } from '..';
```
cfsnyder: Bad import

src/sdk/model/edge/static-routing/__json__/static-route-json.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Interface for Static Route JSON representation
+ */
+export interface StaticRouteJson {
+  edge_uuid: string | null;
```
cfsnyder: should these all actually be | null?

PR #267, opened 2018-06-19
Resolves #265: add endpoints to retrieve and end all active user sess…

### LINKED ISSUE #265: Add endpoint methods for user getSessions and removeAllSessions

bsnyder788:
…ions.


src/sdk/model/user/__json__/user-session-list-json.ts
```diff
@@ -0,0 +1,5 @@
+import { UserSessionJson } from './user-session-json';
+
+export interface UserSessionListJson {
```
cfsnyder: this is never used

bsnyder788:
@cfsnyder removed the extra unused interface

PR #274, opened 2018-06-21
Resolves #264: Add user endpoint method for updating profile pic

### LINKED ISSUE #264: Add user endpoint method for updating profile pic

fradinni:


src/sdk/model/user/user.ts
```diff
@@ -327,4 +328,13 @@ export class User {
     return Iland.getHttp().delete(`/users/${this.username}/sessions`);
   }
 
+  /**
+   * Updates the user's profile picture.
+   * @param {number} pictureBytes
```
JoMas971: The @param is not good. It should be something like : @param {JpegImage} picture

src/sdk/model/user/user.ts
```diff
@@ -327,4 +328,13 @@ export class User {
...
+   * @returns {Promise<void>}
```
JoMas971: It should return a Promise<any>

fradinni: it returns a promise...

fradinni: Iland.get() is a promise

PR #316, opened 2018-07-17
Resolves #315: Fix getRunbookTestDetails return type

### LINKED ISSUE #315: Fix getRunbookTaskDetails return type

nksprzak:


src/sdk/model/org/disaster-recovery-runbook/disaster-recovery-runbook.ts
```diff
@@ -251,11 +253,11 @@ export class DisasterRecoveryRunbook {
    * @param {string} taskUuid
    * @returns {Promise<Task>}
```
cfsnyder: can you update this doc too?

PR #324, opened 2018-07-24
Add the refresh method to the UserWithSecurity class..

JoMas971:


src/sdk/model/user/user-with-security.ts
```diff
@@ -139,4 +142,15 @@ export class UserWithSecurity extends User {
   isPermittedTo(permissionType: PermissionType, entityUuid: string): boolean {
     return IamService.isUserPermitted(this, entityUuid, permissionType);
   }
+
+  /**
+   * Retrieves a new representation of the user from the API.
+   * @returns {Promise<User>} promise that resolves with updated user
+   */
+  async refresh(): Promise<UserWithSecurity> {
+    return Iland.getHttp().get(`/users/${this.username}`).then((response) => {
```
cfsnyder: the body of this should simply be `return super.refresh().then(() => this);`. The other changes in this PR aren't needed.

JoMas971: ho ok I'll.

PR #343, opened 2018-08-15
Add EventList class to get events from filterParams.

JoMas971:


src/sdk/model/event/event-filter-params.ts
```diff
@@ -0,0 +1,126 @@
+import { EntityType } from '../common/__json__/entity-type';
+import { EventFilterParamsJson } from './__json__/event-filter-params-json';
+import { EventType } from './__json__/event-type';
+import { PagingOrder } from '../user/__json__/paging-order';
+
+/**
+ * Class EventFilterParams
+ */
+/* istanbul ignore next: autogenerated */
+export class EventFilterParams {
```
cfsnyder: I would like to see this class be constructed in the same way as the TaskFilterParams class. That is, it should have an overloaded constructor and a method for getting the corresponding query parameters (and you should add an EventFilterQueryParams interface)

cfsnyder:
thanks @JoMas971 !

PR #345, opened 2018-08-21
Add emailEventHistory method to org, vapp and vdc

JoMas971:


src/sdk/model/org/org.ts
```diff
@@ -1800,4 +1800,16 @@ export class Org extends Entity {
     });
   }
 
+  /**
+   * Retrieve a CSV report email with all event history for a given Org.
+   * @returns {Promise<Task>} task promise
+   */
+  /* istanbul ignore next: autogenerated */
+  async emailEventHistory(email: string): Promise<Task> {
```
cfsnyder: did you autogenerate these @JoMas971 ?

JoMas971: no indeed... I've just copy paste it... Will need to write the tests then.

cfsnyder: i ask because the signature is wrong, these endpoints don't return tasks

JoMas971: yeah... I just saw this... The vm signature was already there, so i just copy pasted it on other entity but i didn't see that the signature was wrong. I'm fixing it and add some tests ;-).

cfsnyder: why not just actually generate them using the plugin?

PR #360, opened 2018-08-24
fix ipsec vpn response object

ariquinones:
issue #359


cfsnyder:
thanks @ariquinones !

PR #367, opened 2018-09-05
update firewall update request object

ariquinones:


JoMas971:
Code coverage goes down...
In that case you have 2 solutions :
- create tests for this file to avoid coverage to go down,
- use cory ts generator IntelliJ plugin to generate the ts class from JAVA response class.
It will include `/* istanbul ignore next: autogenerated */` and we won't need to test it SDK side, since it has been tested JAVA side (the ts class should be a perfect reflection of what we have Java side except in some cases...).

ariquinones:
@JoMas971 the request object I updated as well as the method used still have the /* istanbul ignore next: autogenerated */ comment so I'm wondering what file is bringing the coverage down? By updating the rules, I updated the other firewall.test.unit file and it tests all methods within this class, so what file or class do i need to create the test for so coverage doesn't go down?

src/sdk/model/edge/firewall/edge-firewall-update-request.ts
```diff
@@ -10,70 +11,81 @@ export class EdgeFirewallUpdateRequest {
...
+     */
   get defaultAction(): string {
     return this._json.default_action;
   }
 
-  /**
-   * Get the json representation of this class.
-   * @returns {EdgeFirewallUpdateRequestJson}
-   */
+    /**
```
JoMas971: Wrong indentation in this file.

src/sdk/model/edge/firewall/edge-firewall-rule-update-request.ts
```diff
@@ -0,0 +1,194 @@
...
+            // Json or empty constructor
+      this._json = (firstParam || {}) as EdgeFirewallRuleUpdateRequestJson;
+    }
+  }
+
+    /**
+     * Get description.
+     * @returns {string}
+     */
+  get description(): string {
```
JoMas971: Wrong indentation in this file.

JoMas971:
I just have 2 little comments that need to be fixed. Otherwise, the code looks good. The coverage goes dow by only 0.02% which is negligible and caused by the new file addition. It consider it as a change when it shouldn't...

PR #378, opened 2018-09-18
edge static routing service

ariquinones:


cfsnyder:
Is it just that the serialization format changed from v0.9 to v1.0? If so, why aren't you just fixing the serialization server side?

PR #381, opened 2018-09-24
Make `recovery_journal_used_storage` more consistent with other storage properties using MB vs GB at the VPG level.

anguenot:


anguenot:
API has not been deployed to prod which explains why the check is failing.

cfsnyder:
Yep, will merge once the server-side PR is merged @anguenot

PR #384, opened 2018-10-15
Implement role impersonate.

JoMas971:


src/sdk/model/iam/role/__json__/role-json.ts
```diff
@@ -1,6 +1,16 @@
 import { PolicyJson } from '../../policy/__json__/policy-json';
 import { RoleType } from './role-type';
 
+/**
+ * Interface for impersonated role.
+ */
+export interface ImpersonatedRole {
```
cfsnyder: there is no need for this, the only field that is ever used is the role uuid

src/sdk/iland.ts
```diff
@@ -3,20 +3,25 @@ import { Logger } from './logger/logger';
...
+  public static impersonatedRole: ImpersonatedRole | undefined;
+
   private static _authProvider: AuthProvider | undefined;
 
   private static _logger: Logger;
 
   private static _http: Http | undefined;
 
-  static init(_authProvider: AuthProvider, _config?: IlandSdkConfig, _logger?: Logger) {
+  static init(_authProvider: AuthProvider, _config?: IlandSdkConfig, _logger?: Logger,
```
cfsnyder: this should not be passed in as global SDK configuration, instead you should simply have methods on the IlandBrowserAuthProvider for `testRole(roleUuid: string)` and `endRoleTest()` ( no need to put this on the direct grant one right now)

src/sdk/auth/auth-provider-service.ts
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * AuthProviderService class
+ */
+export class AuthProviderService {
+  /**
+   * If we have an impersonated role initialized with the SDK, we need to update all token.
+   * @param token
+   * @returns {string} the complete token.
+   */
+  static getImpersonateToken(token?: string): string {
```
cfsnyder: just make this a private static function within the IlandBrowserAuthProvider

src/sdk/auth/browser-auth-provider.ts
```diff
@@ -37,17 +38,25 @@ export class IlandBrowserAuthProvider implements AuthProvider {
...
       });
     });
   }
 
+  /**
+   * Return the current complete access token
+   * @returns {string} the access token
+   */
+  getTokenSync(): string {
+    return AuthProviderService.getImpersonateToken(this._keycloak.token);
```
cfsnyder: call this function from everywhere else internally so that we're only calling the augmentation function in one place

src/sdk/iland.ts
JoMas971: ok for the testRole() method, but since we are refreshing the whole app do we need a endRoleTest method ? we won't use use it.

cfsnyder: the refresh is an implementation detail of the console UI and is not related to the implementation of the underlying feature in this SDK

JoMas971: ok then. i'll add it though.

src/sdk/auth/browser-auth-provider.ts
JoMas971: i don't understand what you mean there. can you be more specific please ? :-/

src/sdk/auth/auth-provider-service.ts
JoMas971: ok i'll.

src/sdk/auth/browser-auth-provider.ts
cfsnyder: i mean call getTokenSync internally here vs. using getImpersonateToken all over the place

JoMas971: ho !!!! ok i see :-)

src/sdk/model/iam/role/__json__/role-json.ts
cfsnyder: you still haven't removed this @JoMas971

src/sdk/auth/browser-auth-provider.ts
```diff
@@ -93,6 +117,15 @@ export class IlandBrowserAuthProvider implements AuthProvider {
...
     });
   }
+
+  /**
+   * If we have an impersonated role initialized with the SDK, we need to update all token.
+   * @param token
+   * @returns {string} the complete token.
+   */
+  private getImpersonateToken(token?: string): string {
+    return this._testingRoleUuid ? `impersonate=${this._testingRoleUuid};${token || ''}` : token || '';
```
cfsnyder: if token is falsy, you should just return token right away vs. using the || ''

JoMas971: ho ok. so the function can return a `string` or `undefined` actually right ?

cfsnyder: yeah if the token itself can be undefined then the result of this function needs to be able to be undefined as well

JoMas971: Alright. it's now the case ;-). i've just updated the PR.

PR #406, opened 2019-03-08
add vm configuration limits endpoint

ariquinones:


src/sdk/model/location/vm-configuration-limits.ts
```diff
@@ -0,0 +1,63 @@
+import { VmConfigurationLimitsJson } from './__json__/vm-configuration-limits-json';
+
+export class VmConfigurationLimits {
```
cfsnyder: did you not autogenerate this @ariquinones ?

ariquinones: @cfsnyder okay i just updated with the auto generator

PR #408, opened 2019-03-12
fix lint errors

ariquinones:


src/sdk/service/upload-service/upload-service.ts
```diff
@@ -188,6 +188,7 @@ export abstract class UploadService {
           });
         });
       }
+      /* tslint:disable-next-line */
```
cfsnyder: what did you need to disable on this line @ariquinones ?

ariquinones: it was a lint error as well saying you don't use that variable

ariquinones: const client = this.getHTTPClient();

cfsnyder: ok, we just need to remove the line vs. ignore it then. (and thanks for fixing my lint errors)

ariquinones: Okay updated

cfsnyder:
Thanks @ariquinones !

PR #413, opened 2019-04-01
Add nested_hypervisor to VM model

ariquinones:


ariquinones:
@cfsnyder you're talking about the Task.ts changes right? All i did was move that private static method up to the top of the file because it wasn't passing the lint test, I think codecov didn't catch it at first when running the tests

cfsnyder:
Ah ok, thanks @ariquinones . I think I still need to add the lint to the build. I'll do that now.

PR #416, opened 2019-04-29
update virtual disk request for vm reconfiguration

ariquinones:


cfsnyder:
VirtualDiskJson only coincidentally has the same representation as the disk config request object for the vm reconfiguration endpoint. i.e. they are different models server-side. You should generate a new/separate TS interface for that model using the IDEA plugin.

cfsnyder:
@ariquinones  please rebase on master to fix the tests

ariquinones:
@cfsnyder there it is, I had to force-push it should be good now

cfsnyder:
thanks @ariquinones !

PR #437, opened 2019-07-03
Resolves #432: Add new GET firewall endpoint and associated model

### LINKED ISSUE #432: Add new GET firewall endpoint and associated model
We have a new endpoint to add for retrieving an edge firewall service. (this new endpoint will supersede the previous one)

fradinni:


src/sdk/model/edge/edge.ts
```diff
@@ -218,32 +220,45 @@ export class Edge extends Entity {
 
   /**
    * Get the Edge firewall
+   * @deprecated This method is deprecated, use getEdgeFirewall() instead
    * @returns {Promise<Firewall>}
    */
-  async getFirewall(): Promise<Firewall> {
+  async getFirewall(): Promise<EdgeFirewall> {
```
cfsnyder: @fradinni so `getFirewall` returns `EdgeFirewall` and `getEdgeFirewall` returns `Firewall`? Doesn't that seem backwards?

src/sdk/model/edge/firewall/deprecated/edge-firewall-rule.ts
```diff
@@ -0,0 +1,155 @@
+import { EdgeFirewallRuleJson } from './__json__/edge-firewall-rule-json';
+
+/**
+ * FirewallRule class
+ */
+export class EdgeFirewallRule {
```
cfsnyder: did you generate these with the IDEA plugin or write them manually?

fradinni: This class is the old one ( everything that is in /deprecated/folder is old classes ). I'm not sure this one was generated with the plugin i just moved and eventually renamed those old classes.

All new classes are generated with IDEA plugin.

src/sdk/model/edge/edge.ts
cfsnyder: still wondering about this @fradinni ?

fradinni:
@cfsnyder updated

cfsnyder:
thanks @fradinni

PR #445, opened 2019-08-02
Firewall endpoint fixes + new endpoints

fradinni:
Add get firewall rules information endpoint


src/sdk/model/edge/edge.ts
```diff
@@ -777,8 +777,62 @@ export class Edge extends Entity {
...
+      return new EdgeGatewayFirewallObjectList(json);
+    });
+  }
+
+  /**
+   * Retrieves name/ID mappings for all source and destination objects that are assigned to current firewall rules.
+   * @returns {Promise<Array<{name: string, id: string}>>}
+   */
+  /* istanbul ignore next: autogenerated */
+  async getFirewallRulesInformation(): Promise<Array<{name: string, id: string}>> {
```
cfsnyder: Let's give the response a formal model like we do everywhere else

fradinni: OK, let me change that

PR #471, opened 2019-11-08
#465: DHCP - Add subnetMask field to ip pool object

fradinni:


fradinni:
@cfsnyder could you check this PR when you get a chance.
- Added the subnet mask field in IP Pool object.
- Fixed a test line in permission service unit tests.

PR #476, opened 2019-11-15
IP-3725: Add network validation service

r-chowdhury:


SuperConfuserUser:
The first failing [check](https://circleci.com/gh/ilanddev/javascript-sdk/1403?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link) can be fixed by updating the `permissionService.permission.size` in the test to the received `91`. It's a basic test on the expected number of different permission types. VDI added 3 more permission types. DHCP was PRed before that commit, but merged after, so it reverted back to the original 88. Had to update this in my JS-SDK branch too.

SuperConfuserUser:
Is the IDE/linter working ok and showing warnings? There were still quite a bit of TypeScript errors in the main file.

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
+import { Range, NetworkParams } from './network-validation-interface';
+
+/**
+ * Network Validation Service
+ */
+
+export class NetworkValidationService {
+  private static instance: NetworkValidationService;
+
+  static getInstance(): NetworkValidationService {
```
SuperConfuserUser: needs documentation

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+    }
+    return combined;
+  }
+
+  /**
+   * Gets prefix size of net mask.
+   * @param {string} netmask
+   * @return {number} prefix size of netmask
+   */
+
```
SuperConfuserUser: remove extra line

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+        return i;
+      }
+    }
+    return null;
+  }
+
+  /**
+   * Indicates whether a string represents a valid IPv4 address.
+   * @param: ip Address string
+   * @returns boolean
```
SuperConfuserUser: TS doc format for params and return

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+      }
+      return true;
+    }
+    return false;
+  }
+
+  /**
+   * Indicates whether a string represents a valid IPv4 address. Excludes reserved addresses.
+   * @param ipAddress: ip Address string
+   * @returns boolean
```
SuperConfuserUser: TS doc format for param and return

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+    }
+    return subnetIp + '/' + prefixSize;
+  }
+
+  /**
+   * Gets the range of a subnet from the netmask and any IP address on the
+   * network. If either the netmask or IP address are invalid, returns null.
+   * @param {string} netmask the netmask string
+   * @param {string} ipAddress the IP address string
+   * @return {object} an object of the form { start: "x.x.x.x", end: "x.x.x.x" },
```
SuperConfuserUser: returns `Range` type

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+    const testIpInt = this.getIpAsInteger(testIpAddress);
+    return ((netmaskInt & netIpInt) === (netmaskInt & testIpInt));
+  }
+
+  /**
+   * Indicates whether an IP address is in a range of IP addresses. Returns null
+   * if any of the parameters do not represent valid network identifiers.
+   * @param {string} ipAddress the IP address which should be tested for whether
+   *          it is in the range
+   * @param {object} range an object of the form { start: "x.x.x.x", end:
```
SuperConfuserUser: update type

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+      }
+      return true;
+    }
+    return false;
+  }
+
+  /**
+   * Takes a set of IP addresses and a set of IP ranges, and indicates whether
+   * any of the IP addresses belong to any of the ranges.
+   * @param {array} ips of IP strings
```
SuperConfuserUser: Typescript way of typing arrays would be Array<string> or string[]

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+   * @param {array} ranges of range objects
```
SuperConfuserUser: Typescript way of typing arrays would be Array<string> or string[] just like how they're specified in the function parameters.

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Indicates whether any IP in an array of IPs is within an IP range.
+   * @param {array} ips IP address strings
+   * @param {object} range IP range object
```
SuperConfuserUser: update types

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+    }
+    return components.join('.');
+  }
+
+  /**
+   * Parses string to int then checks if test number is between them.
+   * @param {string} a: first number for comparison.
+   * @param {string} b: second number for comparison.
+   * @param test: boolean
+   */
```
SuperConfuserUser: needs a return type

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+    return (port >= 1 && port <= 65535);
+  }
+
+  /**
+   * Gets the number of IP addresses in a given range. Although unlikely to be
+   * an issue, note that due to the fact that javascript numbers are represented
+   * as signed, two's complement integers, for a network with more than 2^31
+   * addresses this function will return a negative number (May not even be
+   * technically possible). Returns -1 if the parameters are invalid.
+   * @param {object} ipRange object representing IP range of the form { start:
```
SuperConfuserUser: update type

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+    }
+    return this.getIpAsInteger(range.end) - this.getIpAsInteger(range.start) + 1;
+  }
+
+  /**
+   * Returns an available IP address within one of an array of given ranges,
+   * given an array of the IP addresses that are already allocated. Returns null
+   * if no addresses are unallocated.
+   * @param {array} ipRanges array of IP ranges to choose from
+   * @param {array} allocatedIpAddresses array of IP addresses which have
```
SuperConfuserUser: update types

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+        return this.getIntegerAsIp(start);
+      }
+    }
+    return null;
+  }
+
+  /**
+   * Returns if every address within range is in ip list
+   * @param {object} ipRanges array of IP ranges to choose from
+   * @param {array} ipList array of IP addresses
```
SuperConfuserUser: update types

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+      }
+    }
+    return true;
+  }
+
+  /**
+   * Indicates whether an IP address is within any range, among a given array of
+   * IP ranges.
+   * @param {string} ip IP address string
+   * @param {array} ranges an array of IP address range objects
```
SuperConfuserUser: update type

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
+import { Range, NetworkParams } from './network-validation-interface';
+
+/**
+ * Network Validation Service
+ */
+
```
SuperConfuserUser: remove extra line

src/sdk/service/network-validation-service/network-validation-interface.ts
```diff
@@ -0,0 +1,9 @@
+export interface Range {
+  start: string;
+  end: string;
+}
+
+export interface NetworkParams {
```
SuperConfuserUser: can have documentation for both the interfaces

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+    const gatewayInt = this.getIpAsInteger(gateway);
+    const maskSize = Number(cidrComponents[1]);
+    const revMaskInt = (2 ** (32 - maskSize)) - 1;
+    return ((gatewayInt & revMaskInt) === 0);
+  }
+
+  /**
+   * Gets the netmask and gateway/ip from a CIDR formatted network string.
+   * @param {string} cidr
+   * @return {object} of the form { gateway: '10.10.10.0', netmask:
```
SuperConfuserUser: update type

r-chowdhury:
> Is the IDE/linter working ok and showing warnings? There were still quite a bit of TypeScript errors in the main file.



> Is the IDE/linter working ok and showing warnings? There were still quite a bit of TypeScript errors in the main file.

image

What error are you getting?

SuperConfuserUser:
> What error are you getting?

Seeing these in the IDE. They also show up in the error logs when trying to build `npm run build`.

Screen Shot 2019-11-19 at 9 33 06 AM

@r-chowdhury

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+      }
+      if (!allocInts.includes(start) && start === end) {
+        return this.getIntegerAsIp(start);
+      }
+    }
+  }
+
+  /**
+   * Returns if every address within range is in ip list.
+   * @param {Range[]} ipRanges: array of IP ranges to choose from.
```
SuperConfuserUser: docs don't match actual param

SuperConfuserUser:
The `isValidMacAddress` function from the JS version of the service seems to be missing.

src/sdk/service/network-validation-service/__tests__/network-validation-service.test.unit.ts
```diff
@@ -0,0 +1,391 @@
...
+  start: '10.0.0.10',
+  end: '10.0.0.255'
+};
+
+test('IP Address is valid', () => {
+  expect(nvs.isValid(dummyIpOne)).toBeTruthy();
+});
+
+test('IP Address is valid and not reserved.', () => {
+  expect(nvs.isValidStrict(dummyIpOne)).toBeFalsy();
```
SuperConfuserUser: Expecting `Falsy` here, so the test description should be updated from `valid` to `invalid`? Same for the rest of the expected `Falsy` tests that are described as `valid`.

r-chowdhury: @SuperConfuserUser I'm testing for both truthiness and falsiness. Plus Cory wanted me to keep the function names the same as the legacy functions.

cfsnyder: I don't recall asking you to keep the same names. I actually would prefer that you change the names where it makes sense. This case is a good example, I would expect `isValidStrict` and `isValid` to be combined. The name could be changed to isValidIPV4Address and it could take an optional boolean parameter to indicate whether .0 and .255 should be considered valid.

cfsnyder: And I agree with Chely's comment about the test description not accurately describing what is being tested here

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+    }
+    return NetworkValidationService.instance;
+  }
+
+  /**
+   * Gets the integer representation of an IP address.
+   * @param {string} ip: the ip address.
+   * @return {number} The integer representation.
+   */
+  getIpAsInteger(ipAddress: string): number | null {
```
cfsnyder: all method names should include IPv4, if relevant because we'll be adding similar methods for IPv6 in the near future and it's better to be explicit

src/sdk/service/network-validation-service/network-validation-interface.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Object representing range of ip addresses.
+ */
+export interface Range {
```
cfsnyder: IPv4AddressRange would be a better, more explicit name

src/sdk/service/network-validation-service/network-validation-interface.ts
```diff
@@ -0,0 +1,15 @@
...
+  start: string;
+  end: string;
+}
+
+/**
+ * Object representing parameters for a network.
+ */
+export interface NetworkParams {
```
cfsnyder: IPv4NetworkDescriptor would be a better, more explicit name

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+/**
+ * Network Validation Service
+ */
+export class NetworkValidationService {
+  private static instance: NetworkValidationService;
+
+  /**
+   * @return {NetworkValidationService} instance of NetworkValidationService class.
+   */
+  static getInstance(): NetworkValidationService {
```
cfsnyder: there is no reason to use a singleton instance for this service. You should just make the class abstract and all methods static.

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+  }
+
+  /**
+   * Indicates whether an IP address is within any range, among a given array of
+   * IP ranges.
+   * @param {string} ip: IP address string.
+   * @param {Range[]} ranges: an array of IP address range objects.
+   * @return {boolean} boolean value.
+   */
+  isIpInAnyRange(ip: string, ranges: Range[]): boolean {
```
cfsnyder: we have a precedent of using Array<?> notation for expressing array types. please update those everywhere.

src/sdk/service/network-validation-service/__tests__/network-validation-service.test.unit.ts
r-chowdhury: Hi Cory. Could you give me an example of what the descriptions should be for some of these tests?

@cfsnyder

r-chowdhury: Actually scratch that. I can probably just look at other tests.

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+        if (component === '' || isNaN(parsedComponent) || parsedComponent % 1 !== 0 ||
+          parsedComponent < 0 || parsedComponent > 255) { return false; }
+      }
+      return true;
+    }
+    return false;
+  }
+
+  /**
+   * Indicates whether a string represents a valid IPv4 address.
```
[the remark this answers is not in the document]
r-chowdhury: I deleted `isValid` all together.

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+   * as signed, two's complement integers, for a network with more than 2^31
+   * addresses this function will return a negative number (May not even be
+   * technically possible). Returns -1 if the parameters are invalid.
+   * @param {Range} ipRange object representing IP range of the form { start:
+   *          "x.x.x.x", end: "x.x.x.x" }
+   * @return {integer} size of range.
+   */
+  getNumberOfAddressesInRange(range: Range): number | null {
+    if (!(this.isValid(range.start) && this.isValid(range.end))) {
+      return null;
```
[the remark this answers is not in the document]
r-chowdhury: I based this on the network-validation-service in the console code, and it returned `null`. @cfsnyder, what do you think?

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,475 @@
...
+  /**
+   * Indicates whether an IP address is in a range of IP addresses. Returns null
+   * if any of the parameters do not represent valid network identifiers.
+   * @param {string} ipAddress: the IP address which should be tested for whether
+   *          it is in the range.
+   * @param {Range} range: an object of the form { start: "x.x.x.x", end:
+   *          "x.x.x.x" }, representing the range of IP addresses.
+   * @return {boolean} boolean value
+   */
+  isInRange(ipAddress: string, range: Range): boolean {
```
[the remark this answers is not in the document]
r-chowdhury: Sorry, I'm not sure I'm following.

src/sdk/service/network-validation-service/__tests__/network-validation-service.test.unit.ts
```diff
@@ -0,0 +1,388 @@
...
+
+test('Check if function fails to get subnet IPv4 Address range.', () => {
+  expect(NetworkValidationService.getSubnetIPv4AddressRange('255.255.192.0', dummyIpThree)).toEqual(
+    {
+      end: '192.168.63.255',
+      start: '192.168.0.0'
+    };
+});
+
+test('Check if function fails to get subnet IPv4 Address range.', () => {
```
cfsnyder: the description here is identical to the previous one. You should be putting multiple `expect` statements within each test case, where appropriate. And for these types of tests, it's usually appropriate to be testing multiple values in each test case to ensure you're covering all edge cases.

src/sdk/service/network-validation-service/__tests__/network-validation-service.test.unit.ts
```diff
@@ -0,0 +1,388 @@
...
+test('Check if not in subnet.', () => {
+  expect(NetworkValidationService.isInSubnet(
+    dummyNetmaskTwo,
+    '192.0.2.130',
+    '192.1.2.130'
+  )).toBeFalsy();
+});
+
+test('Check if invalid netmask or IPv4 address.', () => {
+  expect(NetworkValidationService.isInSubnet(
```
cfsnyder: is there a reason you didn't put IPv4 in the name of this `isInSubnet` method?

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,468 @@
+import { IPv4AddressRange, IPv4NetworkDescriptor } from './network-validation-interface';
+
+/**
+ * Network Validation Service
+ */
+export abstract class NetworkValidationService {
+  /**
+   * Gets the integer representation of an IP address.
+   * @param {string} ip: the ip address.
+   * @return {number} The integer representation.
```
cfsnyder: be explicit in your docs about the fact that this method returns null when the input does not represent a valid IPv4 address

cfsnyder: same comment applies to many of the methods in here

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,468 @@
...
+   */
+  public static getIPv4AsInteger(ipAddress: string): number | null {
```
cfsnyder: This should be private since the representation of an IP as an integer is an implementation detail of the service and does not conform to any kind of specification.

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,468 @@
...
+  }
+
+  /**
+   * Gets prefix size of net mask.
+   * @param {string} netmask: netmask string.
+   * @return {number} prefix size of netmask.
+   */
+  public static getNetmaskPrefixSize(netmask: string): number | null {
+    if (!this.isValidNetmask(netmask)) {
+      return -1;
```
cfsnyder: why should this return -1? It should be documented if there is a good reason.

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,468 @@
+import { IPv4AddressRange, IPv4NetworkDescriptor } from './network-validation-interface';
+
+/**
+ * Network Validation Service
+ */
+export abstract class NetworkValidationService {
+  /**
+   * Gets the integer representation of an IP address.
+   * @param {string} ip: the ip address.
```
cfsnyder: there should not be a colon after the param name. comment applies throughout the PR

src/sdk/service/network-validation-service/__tests__/network-validation-service.test.unit.ts
```diff
@@ -0,0 +1,388 @@
...
+
+test('Check if Netmask is valid.', () => {
+  expect(NetworkValidationService.isValidNetmask('255.255.255.128')).toBeTruthy();
+});
+
+test('Check if Netmask is invalid.', () => {
+  expect(NetworkValidationService.isValidNetmask('255.255.0.128')).toBeFalsy();
+});
+
+test('Check if Netmask is invalid.', () => {
```
cfsnyder: should be combined with the previous one (one test case, multiple assertions). Also, I would expect to see more assertions for both thevalid and invalid cases. i.e. a random string, a netmask with only 3 octets, an empty string, etc. It takes so little time to add an extra assertion that you have no reason to not test things more thoroughly.

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,468 @@
...
+    const gatewayInt = this.getIPv4AsInteger(gateway);
+    const maskSize = Number(cidrComponents[1]);
+    const revMaskInt = (2 ** (32 - maskSize)) - 1;
+    return ((gatewayInt & revMaskInt) === 0);
+  }
+
+  /**
+   * Gets the netmask and gateway/ip from a CIDR formatted network string.
+   * @param {string} cidr: cidr string.
+   * @return {IPv4NetworkDescriptor} of the form { gateway: '10.10.10.0', netmask:
```
cfsnyder: no need to specify the form of the return argument since we're using a statically typed return value.

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,468 @@
...
+    return {
+      gateway: gateway,
+      netmask: netmask
+    };
+  }
+
+  /**
+   * Gets an IP range object from a string of the format: a.b.c.d - e.f.g.h
+   * @param {string} rangeStr: the range string to parse
+   * @return {IPv4AddressRange} range object of the form stated in the description or null if invalid.
```
cfsnyder: again, no need to describe the form of a params in doc comments

cfsnyder: and comment applies throughout PR

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,468 @@
...
+
+  /**
+   * Gets a string representation of an IP address range without the .0 and .255 if applicable. Returns null if either
+   * the netmask or ipAddress are invalid.
+   * @param {string} netmask: the netmask string.
+   * @param {string} ipAddress: the IP address string.
+   * @return {string} a string representation of the IP address range. For
+   *         example, "192.168.0.1 - 192.168.0.10".
+   */
+  public static getSubnetIPv4AddressRangeStrWithoutReserved(netmask: string, ipAddress: string): string | null {
```
cfsnyder: it would be better to combine this and the preceding method and simply include a boolean value as the last param that indicates whether reserved values should be included

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,468 @@
...
+  }
+
+  /**
+   * Takes a set of IP addresses and a set of IP ranges, and indicates whether
+   * any of the IP addresses belong to any of the ranges.
+   * @param {string[]} ips: array of IP strings
+   * @param {IPv4AddressRange[]} ranges: array of range objects
+   * @return {boolean} boolean value
+   */
+  public static isAnyIPv4AddressInAnyIPv4AddressRange(ips: string[], ranges: Array<IPv4AddressRange>): boolean {
```
cfsnyder: I notice a common pattern throughout all methods. i.e. we're constantly dealing with the question of whether the input IP addresses are valid. It would be better for us to create an IPv4Address type that each of these methods could take as input (vs. a free-form string). You could define the IPv4Address type such that it can only be constructed with a valid IPv4 address string. By doing that, you would no longer need to handle the case where the input IP address(es) are invalid within all of these utilities.

src/sdk/service/network-validation-service/network-validation-service.ts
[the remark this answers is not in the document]
cfsnyder: answered by previous comments. namely, that input should be an IPv4Address type so that you aren't dealing with validation of params within each utility method.

src/sdk/service/network-validation-service/network-validation-service.ts
r-chowdhury: Would a hyphen be better or should I literally put nothing? @cfsnyder

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,333 @@
...
+          end: end.address
+        };
+      }
+    }
+    return null;
+  }
+
+  /**
+   * Takes a netmask and any IP address on the network and converts it to a CIDR
+   * string format. If either the netmask or IP address are invalid, returns
```
cfsnyder: Neither the netmask or IP address can be invalid anymore

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,333 @@
...
+      if (i !== 3) {
+        subnetIp += '.';
+      }
+    }
+    return subnetIp + '/' + prefixSize;
+  }
+
+  /**
+   * Gets the range of a subnet from the netmask and any IP address on the
+   * network. If either the netmask or IP address are invalid, returns null.
```
cfsnyder: Comment about returning null isn't relevant

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,333 @@
...
+   * @param {string} netmask: the netmask string.
+   * @param {string} ipAddress: the IP address string.
+   * @return {IPv4AddressRange} an object of the form { start: "x.x.x.x", end: "x.x.x.x" },
+   *         where the start field contains the IP address at the low end of the
+   *         range and the end field contains the IP address at the high end of
+   *         the range.
+   */
+  public static getSubnetIPv4AddressRange(netmask: Netmask, ipAddress: IPv4Address): IPv4AddressRange | null {
```
cfsnyder: type union with null for the return type is no longer valid. Check the docs and return types of all methods

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,333 @@
...
+      if (this.isAnyIPv4AddressInIPv4AddressRange(ips, ranges[i])) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Indicates whether any IP in an array of IPs is within an IP range.
+   * @param {string[]} ips IP address strings.
```
cfsnyder: wrong type

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,333 @@
...
+      return this.isInIPv4AddressRange(ip, range);
+    }).length > 0;
+  }
+
+  /**
+   * Converts integer to ip address representation.
+   * @param {number} ipInt: Integer to be converted.
+   * @return {string} IP Address string.
+   */
+  public static getIntegerAsIPv4Address(ipInt: number): string {
```
cfsnyder: can this be private?

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,333 @@
...
+  }
+
+  /**
+   * Parses string to int then checks if test number is between them.
+   * @param {string} a first number for comparison.
+   * @param {string} b second number for comparison.
+   * @param {string} test number being compared.
+   * @return {boolean}
+   */
+  public static inIPv4AddressRange(a: string, b: string, test: string): boolean {
```
cfsnyder: what is this method for? strange that it takes number args as strings, right?

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,333 @@
...
+   * Returns an available IP address within one of an array of given ranges,
+   * given an array of the IP addresses that are already allocated. Returns null
+   * if no addresses are unallocated.
+   * @param {IPv4AddressRange[]} ipRanges array of IP ranges to choose from.
+   * @param {string[]} allocatedIpAddresses array of IP addresses which have
+   *          already been allocated.
+   * @return {string} an available IP address.
+   */
+  public static getAvailableIPv4(ipv4AddressRanges: Array<IPv4AddressRange>,
+                                 allocatedIpAddresses: IPv4Address[]): string | null {
```
cfsnyder: We use Array<> notation for TS array type declarations in this project. I see you using the [] in some other places too, so please update them all.

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,333 @@
...
+    return null;
+  }
+
+  /**
+   * Returns if every address within range is in ip list.
+   * @param {IPv4AddressRange} range range object.
+   * @param {string[]} ipList array of IP addresses.
+   * @return {boolean}
+   */
+  public static areAllIPv4AddressesFromIPv4AddressRangeInList(ipList: IPv4Address[], range: IPv4AddressRange): boolean {
```
cfsnyder: probably don't need to repeat IPv4Address twice in the method name, I'd remove the second occurrence for the sake of brevity.

src/sdk/service/network-validation-service/network-validation-types.ts
```diff
@@ -0,0 +1,145 @@
...
+      throw new Error('Not valid IPv4 Address.');
+    }
+  }
+
+  /**
+   * Indicates whether a string represents a valid IPv4 address.
+   * @param {string} ipAddress ipAddress string representation.
+   * @returns {boolean}
+   */
+  private isValidIPv4Address(ipAddress: string, strict?: boolean): boolean {
```
cfsnyder: make this public static, will be useful in validators

src/sdk/service/network-validation-service/network-validation-types.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/**
+ * Class for representing CIDR with validations.
+ */
+export class Cidr extends IPv4Address {
+  readonly hostIdentifier: number;
+
+  constructor(cidr: string) {
+    const cidrComponents = cidr.split('/');
```
cfsnyder: refactor the validity check into a public static method on this class

src/sdk/service/network-validation-service/network-validation-types.ts
```diff
@@ -0,0 +1,145 @@
...
+  constructor(address: string) {
+    super(address);
+    if (this.isValidNetmask(address)) {
+      this.netmask = address;
+    } else {
+      throw new Error('Not valid IPv4 Netmask.');
+    }
+  }
+
+  isValidNetmask(netmask: string): boolean {
```
cfsnyder: make static

src/sdk/service/network-validation-service/network-validation-types.ts
```diff
@@ -0,0 +1,145 @@
...
+    } else {
+      this.hostIdentifier = suffix;
+    }
+  }
+}
+
+/**
+ * Class representing IPv4 Netmask with validations.
+ */
+export class Netmask extends IPv4Address {
```
cfsnyder: shouldn't extend IPv4Address, a netmask isn't generally a valid IP

src/sdk/service/network-validation-service/network-validation-types.ts
```diff
@@ -0,0 +1,145 @@
...
+      return true;
+    }
+    return false;
+  }
+}
+
+/**
+ * Class for representing CIDR with validations.
+ */
+export class Cidr extends IPv4Address {
```
cfsnyder: should not extend IPv4Address, a CIDR is not an IP

src/sdk/service/network-validation-service/network-validation-types.ts
```diff
@@ -0,0 +1,145 @@
...
+      throw new Error('Not valid subnet CIDR.');
+    }
+  }
+
+  /**
+   * Checks if string is valid subnet CIDR.
+   * @param {string} cidr cidr string.
+   * @return {boolean}
+   */
+  isValidSubnetCidr(): boolean {
```
cfsnyder: make static

src/sdk/service/network-validation-service/network-validation-types.ts
```diff
@@ -0,0 +1,145 @@
...
+  gateway: string;
+  netmask: string;
+}
+
+/**
+ * Class representing IPv4Address and validation.
+ */
+export class IPv4Address {
+  readonly address: string;
+  readonly strict?: boolean;
```
cfsnyder: this doesn't need to be a field of the class, does it?

src/sdk/service/permission-service/__tests__/permission-service.test.unit.ts
```diff
@@ -35,7 +35,7 @@ function runPermissionServiceViewPermissionAssertions(permission: Permission | u
 
 test('Properly instantiate permission service to get permissions and domain permissions maps', () => {
   const permissionService = PermissionService.getInstance();
-  expect(permissionService.permissions.size).toBe(88);
+  expect(permissionService.permissions.size).toBe(91);
```
cfsnyder: i think we removed this test awhile back, you may need to rebase

tsconfig.json
```diff
@@ -18,7 +18,7 @@
     "noImplicitReturns": true,
     "noImplicitThis": true,
     "lib" : [
-      "es6", "dom"
+      "es6", "dom", "es2017"
```
cfsnyder: do you have a good reason for needing this? If you really want to use es2017, you should do this on a separate PR and do proper due diligence to ensure that it has no unexpected consequences. I also think it would make sense to remove "es6" if you include "es2017", but I'm not sure without looking myself.

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,333 @@
...
+      netmask: netmask.address
+    };
+  }
+
+  /**
+   * Gets an IP range object from a string of the format: a.b.c.d - e.f.g.h
+   * @param {string} rangeStr: the range string to parse
+   * @return {IPv4AddressRange} range object of the form stated in the description or null if invalid.
+   */
+  public static getIPv4AddressRangeFromString(rangeStr: string): IPv4AddressRange | null {
```
cfsnyder: would be nice to have a separate, public method for checking whether a range string is valid and then use that in here.

src/sdk/service/network-validation-service/network-validation-service.ts
```diff
@@ -0,0 +1,333 @@
...
+    }
+    return combined;
+  }
+
+  /**
+   * Gets prefix size of net mask.
+   * @param {string} netmask netmask string.
+   * @return {number} prefix size of netmask.
+   */
+  public static getNetmaskPrefixSize(netmask: IPv4Address): number | null {
```
cfsnyder: never returns null, update return type

src/sdk/service/network-validation-service/__tests__/network-validation-service.test.unit.ts
```diff
@@ -0,0 +1,357 @@
...
+      new IPv4Address('10.0.0.9')
+    ],
+    {
+      start: '10.0.0.1',
+      end: '10.0.0.10'
+    }
+  )).toBeFalsy();
+  expect(NetworkValidationService.areAllIPv4AddressesFromIPv4AddressRangeInList(
+    [
+      new IPv4Address('10.0.0.1'),
```
cfsnyder: FYI, you could be more concise in tests like these by doing it like this:
```
[
  '10.0.0.1',
  '10.0.0.2',
  '10.0.0.3'
].map(s => new IPv4Addres(s))
```

src/sdk/service/network-validation-service/__tests__/network-validation-service.test.unit.ts
```diff
@@ -0,0 +1,357 @@
...
+  );
+  expect(NetworkValidationService.getIPv4AsInteger(new IPv4Address('192.168.0.1'))).toEqual(
+   -1062731775
+  );
+});
+
+
+test('Check if getNetmaskPrefixSize function gets Netmask prefix size.', () => {
+  expect(NetworkValidationService.getNetmaskPrefixSize(new Netmask('255.255.255.0'))).toEqual(
+    24
```
cfsnyder: check a few more, especially the extremes (0, 32). You generally always want to at least test the extreme cases

src/sdk/service/network-validation-service/__tests__/network-validation-service.test.unit.ts
```diff
@@ -0,0 +1,357 @@
...
+    },
+    {
+      start: '10.0.0.11',
+      end: '10.0.0.15'
+    }]
+  )).toBeFalsy();
+});
+
+test('Check if gets IPv4 Address as integer and vice-versa.', () => {
+  expect(NetworkValidationService.getIntegerAsIPv4Address(-1062731775)).toEqual(
```
cfsnyder: I would test a whole list of random IPs for this one and would just test the inversion, not the specific integer values since that is an implementation detail.

i.e.
```
const testIps = [
  '1.2.3.4',
  '2.3.5.6',
  '192.168.0.25',
  '10.10.10.12',
  '1.1.1.1',
  '1.0.0.0',
  '254.254.254.254'
].map(s => new IPv4Address(s));
testIps.forEach(testIp => {
  expect(NetworkValidationService.getIntegerAsIPv4Address(NetworkValidationService.getIPv4AsInteger(testIp)).address).toEqual(testIp.address);
});

(may be typos in the code)

File: package.json [hunk 61]
```diff
@@ -28,14 +28,14 @@
...
-    "@types/jest": "20",
+    "@types/jest": "^20.0.8",
     "@types/node": "7",
     "@types/qs": "6",
     "dts-bundle": "0",
     "gh-pages": "1",
     "glob": "7",
-    "jest": "23",
-    "ts-jest": "22",
+    "jest": "^23.6.0",
```
cfsnyder: typically we'd update these kinds of dependencies on a separate PR and make sure that nothing breaks in the process. What did you need from these updates?

SuperConfuserUser:
> The `isValidMacAddress` function from the JS version of the service seems to be missing.

@r-chowdhury - Was this function not going to be included for the TS version? Still not seeing it.

package.json
r-chowdhury: Nothing. It got updated automatically. I'll change it back.

src/sdk/service/network-validation-service/network-validation-service.ts
r-chowdhury: I don't think so because I'm testing it.

src/sdk/service/network-validation-service/network-validation-service.ts
r-chowdhury: I just blindly copied it from the original. Want me to delete it?

r-chowdhury:
@ariquinones @cfsnyder This is ready for review again. Cory, I left some questions for you as well on how to proceed with some of the issues.

PR #479, opened 2019-12-20
Resolves #470: Add new L2 VPN endpoints

### LINKED ISSUE #470: Add new L2 VPN endpoints

fradinni:


src/sdk/model/edge/certificates/x509-certificate.ts
```diff
@@ -0,0 +1,168 @@
...
+  get notBefore(): number {
+    return this._json.not_before;
+  }
+
+  /**
+   * Get not after.
+   * @returns {number}
+   */
+  get notAfter(): number {
+    return this._json.not_after;
```
cfsnyder: should be a date?

src/sdk/model/edge/certificates/x509-certificate.ts
```diff
@@ -0,0 +1,168 @@
...
+  get signature(): string {
+    return this._json.signature;
+  }
+
+  /**
+   * Get not before.
+   * @returns {number}
+   */
+  get notBefore(): number {
+    return this._json.not_before;
```
cfsnyder: should be a date?

src/sdk/model/iam/permission/__tests__/permission-map.test.unit.ts
```diff
@@ -15,5 +15,5 @@ beforeAll(() => {
 
 test('Properly instantiate a Permission map and get permissions', () => {
```
cfsnyder: remove this test too

src/sdk/model/edge/certificates/x509-certificate.ts
fradinni: yeah we can return a date instead of a timestamp. Good catch.

src/sdk/model/iam/permission/__tests__/permission-map.test.unit.ts
fradinni: ok

PR #488, opened 2020-01-23
Add the ability to reset the password for a tenant.

JoMas971:
Add password reset feature for a tenant.
This PR must be merged before #489


src/sdk/model/vac/ba-company.ts
```diff
@@ -541,4 +543,20 @@ export class BaCompany extends Entity {
     }
     return Iland.getHttp().post(`/vac-companies/${this.uuid}/actions/request-upgrade-contract`, request);
   }
+
+  /**
+   * Reset the password for a given tenant (BaCompany)
+   * @param {BaCompanyPasswordResetRequest | BaCompanyPasswordResetRequestJson} resetPasswordRequest
+   * @returns {Promise}
+   */
+  async resetVacCompanyPassword(resetPasswordRequest: BaCompanyPasswordResetRequest |
```
cfsnyder: why accept both? seems to add unnecessary complexity

JoMas971: hum... indeed you're right. Let me change that quick.

JoMas971: Fixed!

PR #497, opened 2020-02-26
Resolves #486: Add new Microsegmentation endpoints - part 1

### LINKED ISSUE #486: Add new Microsegmentation endpoints - part 1
First segment of the Microsegmentation endpoints includes:
- IP Sets
- MAC Sets
- Applications
- Application Groups

SuperConfuserUser:
Part 1 includes:

- IP Sets
- MAC Sets
- Applications
- Application Groups


JoMas971:
@SuperConfuserUser do you know about @cfsnyder super handy InteliJ plugin that allow us to automatically generate typescript code for stuff like that ? It looks like you didn't use it...
It's better to use it so that we have same code generated everywhere. Also, since it's pretty much a copy paste of code, we don't need to do unit tests on our side since they are done already core side.
Since we are generating code from the (JAVA) API code directly we are 98% sure to have exactly the same code as the API. The remanning 2% are about converting Dates fields or adding some custom SDK methods on our side.
We can have a call about that if you want.

PR #504, opened 2020-03-20
Resolves #503: Handle null objects in L2VPN API responses.

### LINKED ISSUE #503: L2VPN Endpoints - Handle null objects in endpoint responses

fradinni:


src/sdk/model/edge/edge-gateway-l2vpn/edge-gateway-l2vpn-peer-site.ts
```diff
@@ -34,7 +34,10 @@ export class EdgeGatewayL2VpnPeerSite {
    * Get l2 vpn user.
    * @returns {EdgeGatewayL2VpnUser}
    */
-  get l2VpnUser(): EdgeGatewayL2VpnUser {
+  get l2VpnUser(): EdgeGatewayL2VpnUser | undefined {
```
JoMas971: Shouldn't we return null instead ?

ariquinones: Agree with Jon on this one

src/sdk/model/edge/edge-gateway-l2vpn/edge-gateway-l2vpn-peer-site.ts
```diff
@@ -43,6 +46,9 @@ export class EdgeGatewayL2VpnPeerSite {
    * @returns {Array<number>}
    */
   get vnics(): Array<number> {
+    if (!this._json.vnics) {
```
ariquinones: I think all these if return statements are good cases to use the ternary operator, instead of each being 4 lines they'll be one

SuperConfuserUser: It could be simplified even a little more. return this._json.vnics || []

PR #507, opened 2020-03-31
Add requestVacContractUpgrade to company object.

JoMas971:
This is needed for IP-3884.


JoMas971:
@cfsnyder : I have a question about how to deprecate an endpoint SDK side. Because this new endpoint will replace the old one that is bound to the BaCompany object. It may be deprecated API side so i wonder what we should do on our side to deprecate the other endpoint.

JoMas971:
@SuperConfuserUser @ariquinones @fradinni : Can i get a +1 on this one please ?

PR #537, opened 2020-12-01
NSX - Add DFW endpoints

fradinni:


src/sdk/model/vdc/distributed-firewall/__json__/distributed-firewall-layer2-json.ts
```diff
@@ -0,0 +1,7 @@
+import { DistributedFirewallJson } from './distributed-firewall-json';
+
+/**
+ * Distributed Firewall Layer 2 Object JSON (Ethernet).
+ */
+export interface DistributedFirewallLayer2Json extends DistributedFirewallJson {
```
ariquinones, line 6: Extending this interface with what? Or for what purpose?

src/sdk/model/vdc/distributed-firewall/distributed-firewall-layer2-update-request.ts
```diff
@@ -0,0 +1,97 @@
+
+/* istanbul ignore next: autogenerated */
```
ariquinones, line 2: This autogenerated comment should be above the exported class

ariquinones (commented):
Just two small comments @fradinni 
Will let @JoMas971 @SuperConfuserUser @iamtash take a look in case I missed anything

src/sdk/model/vdc/distributed-firewall/__json__/distributed-firewall-layer2-json.ts
fradinni: The idea behind this is:
- We don't know how NSX API and objects will evolve in future, DistributedFirewall Layer2 and Layer3 are different objects with separated endpoints, for the time being objects are using the same parameters (inherited from DistributedFirewallJson) but we don't know if it will remains like that, so if one or more params are added to Layer2 or Layer3 we can add them in these classes.

src/sdk/model/vdc/distributed-firewall/distributed-firewall-layer2-update-request.ts
fradinni: Good catch, thank you.

src/sdk/model/vdc/distributed-firewall/__json__/distributed-firewall-layer2-json.ts
ariquinones: Ah gotcha. Thanks for the explanation @fradinni

ariquinones (approved):
Thanks for the changes Nico, will let @JoMas971 @iamtash @SuperConfuserUser make their comments or approve when ready

PR #539, opened 2020-12-01
Ab beta first commit

SuperConfuserUser:
Adding in current advanced backup work to main repo's ab-beta branch.


ariquinones (changes requested):

ariquinones (commented):
I did a first pass now and found that one comment @SuperConfuserUser. Will let @JoMas971 @fradinni @iamtash take a look before merging this beta branch we'll all use

src/sdk/model/advanced-backups/backup-task/backup-restore-task.ts
```diff
@@ -0,0 +1,126 @@
+import { RestoreTaskStatus } from './__JSON__/restore-task-status-enum';
+import { RestoreTaskType } from './__JSON__/restore-task-type';
+import { BackupRestoreTaskJson } from './__JSON__/backup-restore-task-json';
+import { RestoreTaskObject } from './restore-task-object';
+import { RestoreTaskObjectState } from './restore-task-object-state';
+
+/**
+ * Backup restore task.
+ */
+export class BackupRestoreTask {
```
iamtash, line 10: cory merged [this PR](http://git.iland.test/projects/IPS/repos/svc-cohesity-iaas-backups/pull-requests/38/diff#lib/recovery/recovery-api-base-impl/src/main/java/com/iland/cohesity/iaas/backups/recovery/baseimpl/model/CohesityRestoreTask.java) earlier today. would it be quick and simple enough to add the new `taskName` field to this model here? or would we prefer to wait on this big guy to be merged as is and then separately PR anything else going forward to ab-beta? only asking cuz it's tiny 😅

SuperConfuserUser: Added. But there currently seems to be api side issues when running vdc.listBackupRestoreTasks() before and after model update.

src/sdk/model/advanced-backups/backup-run/__json__/status-backup-run.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Status Backup Run Enum.
+ */
+export enum StatusBackupRun {
```
ariquinones, line 4: Missing WARNING status here

ariquinones (changes requested):

src/sdk/model/advanced-backups/backup-run/__json__/status-backup-run.ts
SuperConfuserUser: added

src/sdk/model/advanced-backups/backup-group/__json__/iaas-backup-subscription-json.ts
```diff
@@ -0,0 +1,17 @@
...
+export interface IaasBackupSubscriptionJson {
+  company_id: string;
+  location_id: string;
+  org_uuid: string;
+  vdc_uuid: string;
+  activation_time: number;
+  deactivation_time?: number;
+  reserved_amount_gb: number;
+  reserved_cost: number;
+  burst_cost_per_gbhour: number;
```
JoMas971, line 15: It should be: 'burst_cost_per_gb_hour'. If it is not the case API side, then they have a JSON formatting issue on their side ;-).

ariquinones (approved):

JoMas971 (commented):
I only have one comment. It looks great otherwise.

src/sdk/model/advanced-backups/backup-group/__json__/iaas-backup-subscription-json.ts
iamtash: it looks like this does match the [API response](http://doc.10.api.iland.test/1.0/#/vDCs/listIaasBackupSubscriptions)

SuperConfuserUser: In this case, should we A) request and wait for an API update or B) merge and update the SDK with separate patch later.

JoMas971: Ok, so this is an issue API side. When it's happening, we need to inform the platform people (i guess in this case it's Cory ? 🤔 ) that they have a JSON format issue. This is a really common issue, it happens. We just need to inform them during the development when we see stuff like that.
In this particular case, i would say, lets create a ticket platform side to ask for the change, and assign it to Cory with a 'Major' priority then create another one SDK side to update it in a future patch (once the fix is deployed platform side) .
That way we can merge this one so @iamtash can PR upon ab-beta branch directly. 😄

JoMas971 (approved):

JoMas971 (changes requested):
Although, it looks like you've impacted the unit test coverage a lot. You need to use `/* istanbul ignore next: autogenerated */` upon all auto generated classes (and maybe enums, i'm not sure). Otherwise, you need to create unit tests.
Reminder: The rule here is simple: If we've used Cory's IntelliJ plugin to generate the typescript code, we don't need unit tests because it is already tested platform side. In any other cases, we need to create the unit tests SDK side. Also, in some cases, we might want to create integration tests. But i'm not sure this is necessary here since everything looks like tested already platform side. But i may be wrong...

src/sdk/model/advanced-backups/backup-group/__json__/iaas-backup-subscription-json.ts
SuperConfuserUser: Sounds good. I'll create an issue for the API platform fix.

SuperConfuserUser:
> Although, it looks like you've impacted the unit test coverage a lot. You need to use `/* istanbul ignore next: autogenerated */` upon all auto generated classes (and maybe enums, i'm not sure). Otherwise, you need to create unit tests.
> Reminder: The rule here is simple: If we've used Cory's IntelliJ plugin to generate the typescript code, we don't need unit tests because it is already tested platform side. In any other cases, we need to create the unit tests SDK side. Also, in some cases, we might want to create integration tests. But i'm not sure this is necessary here since everything looks like tested already platform side. But i may be wrong...

@JoMas971 The TS auto-generate plugin didn't work for a while. So if it doesn't have the `/* instanbul ...*/`, it was hand generated. Then, I also do basic CRUD tests on the SDK to check responses before sharing the SDK commit # for use. The API does have tests Platform side, and most of it has already been implemented Console side as well except for the tickets that we still have left in the epic. Mostly recovery and billing that I can think of.

If unit tests are still required for the sake of code coverage, what is the priority for writing them compared to working on the outstanding AB Console side tickets?

JoMas971:
> > Although, it looks like you've impacted the unit test coverage a lot. You need to use `/* istanbul ignore next: autogenerated */` upon all auto generated classes (and maybe enums, i'm not sure). Otherwise, you need to create unit tests.
> > Reminder: The rule here is simple: If we've used Cory's IntelliJ plugin to generate the typescript code, we don't need unit tests because it is already tested platform side. In any other cases, we need to create the unit tests SDK side. Also, in some cases, we might want to create integration tests. But i'm not sure this is necessary here since everything looks like tested already platform side. But i may be wrong...
> 
> @JoMas971 The TS auto-generate plugin didn't work for a while. So if it doesn't have the `/* instanbul ...*/`, it was hand generated. Then, I also do basic CRUD tests on the SDK to check responses before sharing the SDK commit # for use. The API does have tests Platform side, and most of it has already been implemented Console side as well except for the tickets that we still have left in the epic. Mostly recovery and billing that I can think of.
> 
> If unit tests are still required for the sake of code coverage, what is the priority for writing them compared to working on the outstanding AB Console side tickets?

- Hum, the plugin still works for me (i'm running the latest version of IntelliJ 2020.3.1)... Are you sure you're running the latest version (v 1.1.12) ? 🤔 There is some stuff to do as well, to make it work properly. If you have any question we can figure this together ;-).
- Well, i'll say that for all hand-made classes that are actually already tested platform side, you can go ahead and add the `/* istanbul ignore next: autogenerated */` comment on top of them. That will tell the test coverage to not take them into account. And for the others, you need to add unit tests.

src/sdk/model/advanced-backups/backup-group/__json__/iaas-backup-subscription-json.ts
iamtash: JSON formatting fix was [merged](http://git.iland.test/projects/IC/repos/iland-core/pull-requests/4394/overview)

iamtash:
@JoMas971 please review/merge when you can 😉

SuperConfuserUser:
@JoMas971 - updated a unit test where required and bypassed coverage checks with `/* istanbul ...*/`. Not sure where the diff hits are still coming from. The affected percentage is far better now and much smaller.

JoMas971 (approved):

PR #550, opened 2021-01-15
Advanced backups download bundle link endpoint

ariquinones:


src/sdk/model/vm/vm.ts
```diff
@@ -1295,6 +1295,23 @@ export class Vm extends Entity implements EntityWithPerfSamples {
       return new Blob([response.data], { type: 'application/zip' });
     });
   }
+
+  /**
+   * Get an Advanced Backups download bundle link
+   * @param taskUuid {string} the task uuid used to create bundle
+   * @param fileName {string} optional file name
```
SuperConfuserUser, line 1302: Just double checking since it's new. This is an option that the user can supply as the download's file name?

src/sdk/model/vm/vm.ts
```diff
@@ -1295,6 +1295,23 @@ export class Vm extends Entity implements EntityWithPerfSamples {
...
+   * @returns {Observable<string>}
+   */
+  /* istanbul ignore next: autogenerated */
+  getDownloadBundleLink(taskUuid: string, fileName?: string): Observable<string> {
```
SuperConfuserUser, line 1306: Also, would it be possible to pass the token from the Console in these params, so a subscription wouldn't be necessary? Or would that be weird with 3rd parties using the SDK? Or some other token expiration issues?

src/sdk/model/vm/vm.ts
ariquinones: Okay looks like the content-disposition is set on the server side so I had to take this option off. According to the MDN, if the Content-Disposition header provides a different filename than one outlined in the "download" attribute or "filename", the Content-Disposition will take precedence

src/sdk/model/vm/vm.ts
ariquinones: The token is kept sdk side not console side so calling the sdk to get the token to then pass it in to another sdk method doesn't make sense. And using the observable verifies we'll get the most up to date token, the other method getTokenSync could essentially return undefined

SuperConfuserUser (approved):

SuperConfuserUser:
Thanks for checking and for the explanations!

PR #551, opened 2021-01-21
fix accept header for company.getLogo

iamtash:
currently company logo GET requests are 404-ing bc the accept header is wrong. see [swagger doc](http://doc.10.api.iland.test/1.0/#/Companies/getLogo)

### DIFF
--- a/src/sdk/model/company/company.ts
+++ b/src/sdk/model/company/company.ts
@@ -70,6 +70,7 @@ import { O365AuditLogJson } from './__json__/o365-audit-log-json';
 import { O365AuditLog } from './o365-audit-log';
 import { O365BackupRepositoryJson } from './__json__/o365-backup-repository-json';
 import { O365BackupRepository } from './o365-backup-repository';
+import { Http } from '../../service/http/http';
 
 /**
  * Company.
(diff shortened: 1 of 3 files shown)


src/sdk/model/company/company.ts
```diff
@@ -530,7 +530,7 @@ export class Company extends Entity {
   async getLogo(): Promise<Uint8Array | null> {
     return Iland.getHttp().get(`/companies/${this.uuid}/logo`, {
       headers: {
-        'Accept': 'image/jpeg'
+        'Accept': 'image/vnd.ilandcloud.api.v1.0+jpeg'
```
ariquinones, line 533: Lets go ahead and use the Http.versionMime method for this please

iamtash: oh i see - using now! looks like this is the first use of `Http.versionMime('image/jpeg')` so i added it to the `Http` unit tests as well

ariquinones (approved):

src/sdk/model/company/company.ts
ariquinones: Ah perfect thanks @iamtash !

JoMas971 (approved):

PR #552, opened 2021-01-25
clean up O365 model docs

iamtash:


ariquinones (approved):
Thanks for this @iamtash!

JoMas971 (approved):

PR #553, opened 2021-02-04
Adding Eula sections to template configurations

ariquinones:


JoMas971 (approved):

src/sdk/model/vapp-template/__json__/vapp-template-configuration-json.ts
```diff
@@ -10,4 +11,5 @@ export interface VappTemplateConfigurationJson {
   description: string;
   vms: Array<VmTemplateConfigurationJson>;
   networks: Array<VappTemplateNetworkConfigurationJson>;
+  eula_sections: Array<EulaSectionJson>;
```
SuperConfuserUser, line 14: Since it's a new property on an existing class, this value will never be null or undefined? It should always return at least an empty array?

ariquinones: @SuperConfuserUser yes correct, I double checked and all templates that dont have any eulas will just contain empty arrays for this field including all old templates now too

SuperConfuserUser (approved):

src/sdk/model/vapp-template/__json__/vapp-template-configuration-json.ts
SuperConfuserUser: Thanks for checking! Just wanted to be sure.

PR #555, opened 2021-02-11
add GET integrated backup status endpoints

iamtash:
[platform PR](http://git.iland.test/projects/IC/repos/iland-core/pull-requests/4427/diff#web-v1/rest-api/public-api/src/main/java/com/iland/core/web/rest/response/iaas/intbackup/CompanyLocationIntegratedBackupStatusResponse.java)


ariquinones:
@iamtash looks like you have conflicts now, could you rebase again, I think we should be good after that

iamtash:
thanks @ariquinones - rebased!

src/sdk/model/vdc/__json__/vdc-json.ts
```diff
@@ -23,6 +23,8 @@ export interface VdcJson extends EntityJson {
   used_network_count: number;
   location_id: string;
   org_uuid: string;
+  has_iaas_backups: boolean;
```
JoMas971, line 26: In your [previous](https://github.com/ilanddev/javascript-sdk/pull/556) PR to ` ab-beta`, you're using `has_advanced_backups`, shouldn't it be the same here ? 🤔 
I've checked the swagger for API endpoint and it looks like it returns both `has_advanced_backups` and `has_iaas_backups`. Aren't they the same ? Shouldn't we use only one or the other ?
This PR is against master, if you where using `has_iaas_backups` on the console, after merging `ab-beta` (which will delete this property like in your [previous](https://github.com/ilanddev/javascript-sdk/pull/556) PR) we would have some discrepancy to fix. But if you use `has_advanced_backups` right away we won't.

src/sdk/model/vdc/__mocks__/vdc.ts
```diff
@@ -52,7 +54,9 @@ export const MockSecondVdcJson: VdcJson = {
   name: 'PAYG',
   deleted: false,
   deleted_date: null,
-  updated_date: 1499693404654
+  updated_date: 1499693404654,
+  has_iaas_backups: true,
```
JoMas971, line 58: Same question here.

JoMas971 (changes requested):

src/sdk/model/vdc/__json__/vdc-json.ts
iamtash: @JoMas971 here, `has_iaas_backups` is serving as a new flag that indicates whether a vDC has _either_ of the two types of Iaas backups we offer - Veeam integrated backups _or_ advanced backups. previously, it was used to indicate whether a vDC has the advanced backups offering, but now we have new APIs exposing the Veeam integrated backup status of entities. this PR is to master because integrated backups falls under the general IAAS scope, and i'm keeping the `vdc.has_advanced_backups` flag within the scope of `ab-beta` because it is specific to advanced backups

iamtash: @JoMas971 that said, i'll need ab-beta rebased on master in order to access these new integrated backups endpoints for [this AB issue](http://issues.iland.test/browse/IP-4109), so in ab-beta the vDC model will have all three flags: 
```  
  has_iaas_backups: boolean; // vDC has *either* integrated *or* advanced backups
  has_integrated_backups: boolean; // vDC has the Veeam integrated backups offering
  has_advanced_backups: boolean; // vdc has the advanced backups offering
```

JoMas971 (approved):

ariquinones (approved):

PR #558, opened 2021-02-15
Adding downloadUsers endpoint to O365 org

ariquinones:
The users endpoint for an O365 organization is a paginated one, therefore we have now added the ability to request the entire o365 users together in csv format for user to download.
We have also gotten rid of the repository fields of an o365 user model since those only apply at the Org level


src/sdk/model/o365/o365-organization.ts
```diff
@@ -330,4 +331,23 @@ export class O365Organization extends Entity {
...
+  /* istanbul ignore next: autogenerated */
+  async downloadUsers(downloadProgress?: (e: any) => void): Promise<Blob> {
+    return Iland.getHttp().post(`/o365-organizations/${this.uuid}/users-export`, null, {
+      headers: {
+        'Accept': Http.versionMime('application/octet-stream')
+      },
+      responseType: 'arraybuffer',
+      onDownloadProgress: downloadProgress
+    }).then((response) => {
+      return new Blob([response.data], { type: 'application/csv' });
```
SuperConfuserUser, line 350: type 'text/csv' may be [more](https://stackoverflow.com/questions/18927865/whats-the-difference-between-using-application-csv-vs-text-csv) [appropriate](https://stackoverflow.com/questions/393647/response-content-type-as-csv) here, but 'application/csv' also valid.

JoMas971 (approved):

src/sdk/model/o365/o365-organization.ts
ariquinones: Fixed, thanks @SuperConfuserUser

SuperConfuserUser (approved):

PR #568, opened 2021-03-19
Fix vbo iam permissions

ariquinones:


SuperConfuserUser (approved):

iamtash:
what does required for custom policy mean exactly and how is that determined initially? @ariquinones

iamtash (approved):

ariquinones:
@iamtash to answer your question: what does required for custom policy mean exactly and how is that determined initially? It means that certain permissions need to be implicitly a part of any custom policy for the entity otherwise the custom policy would not work correctly. For example, most view type permissions need to required, in case the custom policy wants to add certain write type permissions, the policy would obviously need the read permission too otherwise the write would not behave as expected. And these permissions are determined initially as a standard, along with certain behaviors the Product team may ask for.

iamtash:
makes sense. thanks for taking the time to explain @ariquinones

PR #577, opened 2021-04-23
add the org dns zone model

iamtash:
adding the model/response actually returned by `Org.getDnsZones()` and fixing the endpoint accordingly.
swagger doc: http://doc.10.api.iland.test/1.0/#/Organizations/getDNSZones

### DIFF
--- a/src/sdk/model/org/org.ts
+++ b/src/sdk/model/org/org.ts
@@ -131,6 +131,8 @@ import { OrgBackupSummaryStats } from '../advanced-backups/backup-run/org-backup
 import { OrgBackupSummaryStatsJson } from '../advanced-backups/backup-run/__json__/org-backup-summary-stats-json';
 import { OrgBackupStatus } from '../advanced-backups/backup-status/org-backup-status';
 import { OrgBackupStatusJson } from '../advanced-backups/backup-status/__json__/org-backup-status-json';
+import { OrgDnsZone } from './org-dns-zone';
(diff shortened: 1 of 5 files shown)


src/sdk/model/org/org.ts
```diff
@@ -514,12 +516,12 @@ export class Org extends Entity {
 
   /**
    * Gets all DNS zones that exist within the org.
-   * @returns {Promise<Array<DnsZone>>} a promise that resolves with the list of DNS zones
+   * @returns {Promise<Array<OrgDnsZone>>} a promise that resolves with the list of DNS zones
    */
-  async getDnsZones(): Promise<Array<DnsZone>> {
+  async getDnsZones(): Promise<Array<OrgDnsZone>> {
```
SuperConfuserUser, line 521: Can rename this to getOrgDnsZones to match the return type update. Is the original DnsZone used by anything else?

iamtash: yes `DnsZone` is returned by `Org.addDnsZone()`

SuperConfuserUser (approved):

src/sdk/model/org/org.ts
SuperConfuserUser: Thanks, I see that this matches the API now.

ariquinones (commented):
Won't merge this yet as it may cause compatibility issues if the console side pr isn't merged into master

JoMas971 (approved):

ariquinones (approved):

PR #591, opened 2021-06-16
Add burst count licenses & total cost for used licenses fields to O365 Bill model

Jchazari:
https://us-east-1.api.ilandcloud.com/doc/current/#/Companies/getO365BillingHistory

### DIFF
--- a/src/sdk/model/company/o365-bill.ts
+++ b/src/sdk/model/company/o365-bill.ts
@@ -11,7 +11,16 @@ export class O365Bill {
   }
 
   /**
+   * Get the number of licenses exceeding the reserved licenses count
+   * @returns {number}
+   */
+  get burstCountLicenses(): number {
+    return this._json.burst_count_licenses;
+  }
+
+  /**
    * Get cost per license of O365 bill
+   * @returns {number}
    */
   get costPerLicense(): number {
     return this._json.cost_per_license;
@@ -19,6 +28,7 @@ export class O365Bill {
 
   /**
    * Get currency code of O365 bill
+   * @returns {CurrencyCode}
    */
   get currencyCode(): CurrencyCode {
     return this._json.currency_code.currency_code;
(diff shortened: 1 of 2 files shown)


src/sdk/model/company/o365-bill.ts
```diff
@@ -10,6 +10,13 @@ export class O365Bill {
   constructor(private _json: O365BillJson) {
   }
 
+  /**
+   * Get the number of licenses exceeding the reserved licenses count
+   */
```
iamtash, line 15: `@returns` doc

src/sdk/model/company/o365-bill.ts
```diff
@@ -52,6 +59,13 @@ export class O365Bill {
     return this._json.total;
   }
 
+  /**
+   * Get the used total including burst usage if any
+   */
```
iamtash, line 64: `@returns` doc

iamtash (changes requested):
'Added' --> 'Add' in the commit message. the general convention is to keep the action/verb (e.g., add, upgrade, fix, remove, refactor) in a commit message  in the present tense.

Juchazari:
@iamtash I should have asked before pushing updates, but there are multiple methods without return doc, should I add them for those methods as well? Also, like you said it should be `@returns` correct? Because there are two methods I see with one having `@return` and the other having `@returns`.

iamtash:
> @iamtash I should have asked before pushing updates, but there are multiple methods without return doc, should I add them for those methods as well? Also, like you said it should be `@returns` correct? Because there are two methods I see with one having `@return` and the other having `@returns`.

pretty sure `@return` and `@returns` are synonyms. both are fine for me, i guess as long as within a particular file the tag usage is consistent. yeah if you don't mind filling in the missing docs that'd be great 🙂

iamtash (approved):

ariquinones (approved):

PR #594, opened 2021-06-25
IP-4271: Add reset 2FA for user method

fradinni:


SuperConfuserUser (commented):
Can document that leaving the roles and role_uuids blank will apply two factor to all users. Either in the settings classes, the company methods, or jsons.

PR #609, opened 2021-08-20
Add teams for O365 backup job

Juchazari:
http://doc.10.api.iland.test/1.0/#/Office%20365%20Backup/getTeams
http://doc.10.api.iland.test/1.0/#/Office%20365%20Backup/getJobByUuid


iamtash (changes requested):
need to add the `type_vet` field to `O365JobJson` and the corresponding getter to `O365Job` no?

Juchazari:
> need to add the `type_vet` field to `O365JobJson` and the corresponding getter to `O365Job` no?

Yes, will actually be adding that for Create Restore Session PR (IP-4399) since it's only really "necessary" for Restore Session / in the scope of that PR.

iamtash:
> > need to add the `type_vet` field to `O365JobJson` and the corresponding getter to `O365Job` no?
> 
> Yes, will actually be adding that for Create Restore Session PR (IP-4399) since it's only really "necessary" for Restore Session / in the scope of that PR.

got it!

PR #678, opened 2022-10-19
Add Vm GuestTools status enums

Juchazari:
enum constants for version status & running status:
https://vdc-download.vmware.com/vmwb-repository/dcr-public/790263bc-bd30-48f1-af12-ed36055d718b/e5f17bfc-ecba-40bf-a04f-376bbb11e811/vim.vm.GuestInfo.ToolsVersionStatus.html

https://vdc-download.vmware.com/vmwb-repository/dcr-public/790263bc-bd30-48f1-af12-ed36055d718b/e5f17bfc-ecba-40bf-a04f-376bbb11e811/vim.vm.GuestInfo.ToolsRunningStatus.html

### DIFF
--- a/src/sdk/model/vm/guest-tools/__mocks__/guest-tools.ts
+++ b/src/sdk/model/vm/guest-tools/__mocks__/guest-tools.ts
@@ -1,9 +1,9 @@
 import { AxiosResponse } from 'axios';
-import { GuestToolsJson } from '../__json__/guest-tools-json';
+import { GuestToolsJson, ToolsRunningStatus, ToolsVersionStatus } from '../__json__/guest-tools-json';
 
 export const MockVmGuestToolsJson: GuestToolsJson = {
-  status: 'fake-status',
-  running_status: 'fake-running-status',
+  status: ToolsVersionStatus.UNMANAGED,
+  running_status: ToolsRunningStatus.NOT_RUNNING,
   version: 'fake-version'
 };
 
--- a/src/sdk/model/vm/guest-tools/__json__/guest-tools-json.ts
+++ b/src/sdk/model/vm/guest-tools/__json__/guest-tools-json.ts
@@ -2,7 +2,31 @@
  * VM Guest Tools JSON
  */
 export interface GuestToolsJson {
-  status: string;
-  running_status: string;
+  status: ToolsVersionStatus;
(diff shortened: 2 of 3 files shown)


src/sdk/model/vm/guest-tools/__mocks__/guest-tools.ts
```diff
@@ -1,7 +1,12 @@
 import { AxiosResponse } from 'axios';
 import { GuestToolsJson } from '../__json__/guest-tools-json';
 
-export const MockVmGuestToolsJson: GuestToolsJson = {
+interface GuestToolsJsonMock extends Omit<GuestToolsJson, 'status' | 'running_status'> {
```
iamtash, line 4: i don't think this is theoretically the best way to handle the TS errors i'm guessing `MockVmGuestToolsJson` is giving you. test data should adhere to the same types real data uses otherwise i imagine the associated tests would be unreliable. i think it would be better to use type assertion on lines 10 and 11. not an expert though so i'll cc @JoMas971

JoMas971: @Juchazari It doesn't make any sense to do what's in `GuestToolsJsonMock`. Because you basically remove 'status' and 'running_status' properties from the first `GuestToolsJson` interface declaration, so you're only picking the 'version' property. Then in `GuestToolsJsonMock` interface declaration you're adding back those two properties removing their original typings and assigning them to string. You're bypassing all type checks there which is bad. 
What you should do instead is removing `GuestToolsJsonMock` interface, and simply use `GuestToolsJson` for your `MockVmGuestToolsJson` const line 9.
Then, you need to use one of the status/version status enums you've created on `src/sdk/model/vm/guest-tools/__json__/guest-tools-json.ts` (pick anyone you like).

You should have something like:
```
export const MockVmGuestToolsJson: GuestToolsJsonMock = {
  status: ToolsVersionStatus. UNMANAGED,
  running_status: ToolsRunningStatus. NOT_RUNNING,
  version: 'fake-version'
};
```

Juchazari: @iamtash @JoMas971 Thank you both! Updated 👍🏼

iamtash (approved):

PR #691, opened 2023-01-03
Changing public value within vapp template

Vinmanftw:
Tested it out on public and non public templates appears to work it now shows false when the template being public is indeed false


Juchazari (commented):
- Change commit message to `change public ...`
- Currently you have 2 commits, need to only have 1. Will need to rebase

Vinmanftw:
Ok sounds good addressing now

Vinmanftw:
Updated @Juchazari

